PR #18, opened 2018-08-20
Fix group wall functions

EraGamer:


EraGamer:
If you'd like me to redo it so that there's only one commit I can.

PR #29, opened 2018-12-22
Add cookieLogin bypass

Neztore:
This new method makes use of Froast's relog function. It's to bypass the login captcha introduced by Roblox recently.

How to use:
- Login to your bot account manually
- Open inspect element (if it's chrome, Control + Shift + i)
- Click "Application"
- Click "Cookies"
- Copy .roblosecurity and run cookieLogin with it

All other sessions for that account will be invalidated and a new token will be stored internally.

After this, all you need to do is run cookieLogin with no parameters. Tokens will be automatically regenerated once a day. Any params will be ignored unless the internal cookie file is out of date.


JonasTheGamer:
How would you use it?

PR #44, opened 2019-02-03
Fixed post.js and getPlayers.js using API

lordhammyrblx:
Fixed post.js using the new [API](https://groups.roblox.com/docs#!/Wall/post_v2_groups_groupId_wall_posts).


lordhammyrblx:
At this point I have finished committing.

PR #56, opened 2019-04-23
Replace URL with /users/profile/profileheader-json

Pokemonjpups:
In my testing, loading the average User Profile can take about 2.5s while loading the profileheader-json only takes around 0.5s or less. This also allows for reading the status of terminated users (although that isn't necessarily the point of this change)

### DIFF
--- a/lib/user/getBlurb.js
+++ b/lib/user/getBlurb.js
@@ -1,6 +1,5 @@
 // Includes
 var http = require('../util/http.js').func
-var parser = require('cheerio')
 
 // Args
 exports.required = ['userId']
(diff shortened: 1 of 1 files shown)


suufi:
Will merge soon with additional changes to other endpoints. Thanks!

chrissie1234:
Is here also an alternative for group information? Like: Group information, current shout, owner of the group and more like that.

suufi:
Okay, so there has been something bugging me making me super apprehensive on merging this and I found it. The code you edited was for `getBlurb.js`, not `getStatus.js`. I'm still going to have to scrape the webpage to get the blurb unless anyone finds an endpoint that actually has the blurb. I will be using this to update other endpoints like `getPlayerInfo.js`.

PR #62, opened 2019-05-05
Provide more error details

chrissie1234:
When this function has an error with an error message, provide more details by adding the error message.


lib/user/follow.js
```diff
@@ -30,7 +30,11 @@ function follow (jar, token, userId) {
           throw new Error(body.message)
         }
       } else {
-        throw new Error('Follow failed')
+        if(body.message){
```
suufi: body is undefined and doesn't conform to standard. please test your code before submitting pull requests

PR #75, opened 2019-07-28
Fix for the errors in exile.js

iAmaury:
I'm not an experienced developer but it seems like this function is currently broken.
I changed something so the string becomes JSON, and now it seems to work.
Before it was just telling me "Cannot read property length of undefined".


iranathan:
why did you close this

iAmaury:
Thought for a second that the fix wasn't working because it was erroring even when I put the correct group and user. But it appears that I put the wrong group ID, oof.

PR #96, opened 2020-01-25
Quality of life improvements

nixpc:
This pull request does 2 things:

1. Gets rid of a "URL is undefined" error by explicitly requiring it.

2. This pull request also uses the most recent cookie instead of the one supplied via cookieLogin. This is done because of the request to signoutfromallsessionsandreauthenticate page which makes the supplied cookie no longer valid and instead stores the new valid cookie on the cookie file.

Now normally we use the cookie file to re-authenticate so during a restart there are no issues but because of the time limit you'll have to constantly provide the bot with new cookies which becomes a real pain for when you want to restart your application.

### DIFF
--- a/lib/util/getVerification.js
+++ b/lib/util/getVerification.js
@@ -3,6 +3,7 @@ var http = require('./http.js').func
 var getHash = require('./getHash.js').func
 var getVerificationInputs = require('./getVerificationInputs.js').func
 var cache = require('../cache')
+var URL = require('url').URL
 
 // Args
 exports.required = ['url']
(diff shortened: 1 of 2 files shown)


suufi:
Hi there! Could you update your code to fit [standard](https://standardjs.com/) formatting?

nixpc:
Hello there @suufi , I'm sorry I didn't get notified of your comment so that's why I'm so late.
I'm not familiar with the standard but I'll read up on it and create a new pull request once I'm done. (Hopefully I don't forget.)

nixpc:
I'll be closing the pull request since I just saw #100 .

PR #113, opened 2020-03-31
Fix shout (Remove JSON.parse)

Neztore:
Not sure whats going on here, but shout has an unnessesary JSON.parse which is causing an error.
Res.body is already an object, not a string.

Simple fix!


Neztore:
This is definitely something to follow up - I've had a brief look through the source and I'm not really sure why it's happening.
Res.body for the `shout` function as returned by `http.js` is an Object:

image

However, res.body for other functions such as getProductInfo is a string. How confusing!

Issue #41, opened 2019-01-27
Custom cookie loading


cloewen8:
I was assigned the task (by Bot Services) of making this dependency compatible with our hosting provider (Heroku). As stated in the README, "we store the cookie internally in a file, and files do not persist in Heroku." I plan to resolve this by storing the cookie in a database.

If I make a pull request to support this, which would be preferable (please add a reaction or comment)?

- **Flexibility**: I add an option to provide functions for saving and loading the cookie. ❤️ 
- **Ease-of-use**: I add an option to provide a connection url to a JDBC-compatible database (requires Java 1.7+). 🚀 
- **Both**: I could do both, whoever in that case, I'd be more inclined to provide *examples* for the save and load functions that are tailored to specific database types (MySQL, SQL Server, PostgreSQL). 👀

Kenner-G:
lmao.

Issue #114, opened 2020-03-31
Discussion: Improving cookie login


Neztore:
It's become clear that an improvement to the current cookie login system is required. Why?
1.  Storing a cookie in JSON on disk doesn't play nicely with services like Heroku
2.  The `cookie` file is **often** unwittingly uploaded to Github by users, putting their accounts at risk.
3.  The current system is not very customisable. You cannot adjust how often tokens are refreshed, or where they are stored.

When I originally implemented this `cookie` system I had not realised the problems that would arise with these services, or how easy it is to just commit all changes and upload them - as many users have done.

Solutions to some of these problems have been suggested in the past, in the form of PRs and issues, see: #41 , #48 and #42 .

I think this is something we need to discuss and try and find a solution for. I have a few possible suggestions:

## Suggestion 1: Do not regenerate tokens by default
This could perhaps involve providing a .regenerate method, or similar.
At present we have no idea how long cookies actually last, and so perhaps responsiblity for rotating tokens often enough should be left to the user?
This means:
- The library is once again stateless, as it should be
- A `regenerate` method, or similar, would use the current jar to generate a new one. It would then update the internally stored `jar` and return the new cookie.

I am most in favour of this solution, though it may make the library less "plug and play" for some users, the cookie file is probably the biggest source of problems.

## Suggestion 2: Make it more customisable
This would involve allowing for values like relogTime to passed, which dictates how often the token should be refreshed - or the passing of a function responsible for storing the token somewhere: Either in Disk or in a database for example. This could be documented to default to storing it in a cookie function.

-------
At any rate, I think this `cookie` behaviour needs to be more fully documented, and we should at least consider improvements even if that is made difficult by Roblox forcing us to use cookies with no guarantee as to how long they last.

chrissie1234:
I agree with suggestion one. Making cookies able to regenerate would be awesome!

JonasTheGamer:
Maybe also a way to "relogin". So for example you could authenticate with cookie1, then do some stuff, and then authenticate with cookie 2, do some suff, and authenticate with cookie1 again :)

AnthonyFuller:
This could possibly be achieved using an enviroment variable and changing that. That will also help with the Heroku issue and it is safer for production.

Issue #48, opened 2019-02-25
Allow configuring cookie file location


LouieK22:
A simple argument for where the cookie file is/what it's named would make deploying apps based on noblox easier.

AnthonyFuller:
You can probably close this now

PR #136, opened 2020-05-28
Proposal: add setCookie & refreshCookie

Neztore:
This PR would revert the changes made to relog to a state where they worked (though they log the user out of all sessions)

It:
- Deprecates cookieLogin
- Adds function setCookie
- Adds function refreshCookie
- Reverts relog.js

# setCookie
All setCookie does is store the provided cookie, and verify it (this is customisable through the novalidate argument)
It does not refresh the cookie.

# refreshCookie
Refreshes either a given cookie or the interally stored cookie, stores the newly generated cookie and returns it. This then places the responsibility for storing the cookie on it's user.
This information will need to be put in docs & readme, of course.

# Why?
See #114 
cookieLogin was intended as a temporary fix and it's been in use since March 2018.
I would say it's *bad* because libraries having unintended effects like this (and of course the creation of a file) makes them more difficult to use and it wreaks havoc with services like Glitch or Heroku. This PR means that these services are supported, provided the user creates their own storage mechanism for the refreshed cookie.

Another improvement about this is that we don't know *when* cookies expire, so the user can decide that themselves. I've personally had a lot of problems because the cookie file is stored relative to the file that runs it, not the library itself - and the cookie file creates an inherent security issue with many users unwittingly uploading their cookie file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,5 @@ jspm_packages/
 .env
 
 \.vscode/
+
+.idea/
--- a/lib/user/cookieLogin.js
+++ b/lib/user/cookieLogin.js
@@ -8,6 +8,7 @@ var interval
 var day = 86400000
 exports.func = function (args) {
   // Run relog
+  console.warn('Noblox - The use of cookieLogin is deprecated: Please make use of the new methods setCookie & refreshCookie.')
   return relog(args.cookie)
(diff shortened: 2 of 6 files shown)


Neztore:
Also closes #41 and #48  though they won't autoclose now

PR #139, opened 2020-05-30
Delete post.js; Captcha

### LINKED ISSUE #138: [BUG] Group/post.js broken (Captcha error and unexpected token o)
**Describe the bug**
group/post.js is completely broken. It first throws error "Unexpected token o in JSON" when parsing, then if you go in and just remove the JSON.parse(), you get a captcha error.

**To Reproduce**
Use the noblox.post() function.

**Expected behavior**
It should not throw any errors and post the given message to the group wall.

Neztore:
There is now a captcha on adding wall posts.
I see no reason to keep this function in place - if Roblox removes the captcha it's code will still be availible in Git history
Fixes #138


Neztore:
Closes #138

PR #153, opened 2020-06-21
Add "get transactions" endpoints & update getPlayers

GeneralScripter:
These endpoints return the list of items sold by a group or user. Will be useful for searching through sales to find or analyse them. The user endpoint can only be used for the currently authenticated user - I pointed out in the commit that it may be best to just use the userId of the said user.

May also be useful to have an onSale/onPurchase event for groups and user. If anyone wants to do that, go ahead. I might if I find time, but I don't specifically need these events at the moment.

Yes, this PR is pretty much just copypasta but I'd like access to these endpoints without using a fork of noblox.

### DIFF
--- a/lib/group/getGroupTransactions.js
+++ b/lib/group/getGroupTransactions.js
@@ -1,0 +1,47 @@
+var http = require('../util/http.js').func
+var getGeneralToken = require('../util/getGeneralToken.js').func
+
+exports.required = ['group']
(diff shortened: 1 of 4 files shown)


GeneralScripter:
Updated getPlayers function to be more reliable and an optional parameter to define a maximum number of users to be returned.

suufi:
Could you please add these to the [typings file](https://github.com/suufi/noblox.js/blob/master/typings/index.d.ts)?

PR #156, opened 2020-07-03
Add getPlayerThumbnail

OptimisticSide:
This endpoint will return a player's thumbnail when given the user-ID of the player, the thumbnail-type of the thumbnail, and the thumbnail size. This will be useful when displaying information about a user. I'd just like to access these endpoints without a fork. (This is my first pull request, so correct me if I'm wrong.)

### DIFF
--- a/lib/user/getPlayerThumbnail.js
+++ b/lib/user/getPlayerThumbnail.js
@@ -1,0 +1,23 @@
+// Includes
+var http = require('../util/http.js').func
+
+// Args
+exports.required = ['userId', 'thumbnailType']
+
+// Define
+exports.func = function (args) {
+  return http({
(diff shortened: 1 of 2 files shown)


suufi:
Could you add this to the typings file?

OptimisticSide:
Ok I will

suufi:
Please test your code. It fails `$ npm test`.

Neztore:
The code now works and passes test, however:
- Is this the best endpoint to use? I believe Roblox deprecated the Roblox.com/* endpoints? (We could thumbnails.roblox.com instead?)
- The typings file does not contain typings for the required values thumbnailType and size.

Neztore:
Hi, another PR has just been merged which implements this functionality. 

Thanks for your PR!

PR #157, opened 2020-07-11
Implement onAuditLog and fix getWall

conman0552:
Support for tracking audit log updates. Functions like all the other group events.


conman0552:
Additionally updated typings file so the blind can utilize this event.

conman0552:
Also resolved this issue as well. https://github.com/suufi/noblox.js/issues/154

PR #160, opened 2020-07-15
Add files via upload

conman0552:
Fixes to remove Milliseconds and solve multiple log occurrences.

### DIFF
--- a/lib/group/onAuditLog.js
+++ b/lib/group/onAuditLog.js
@@ -18,7 +18,7 @@ exports.func = function (args) {
             for (var key in audit.data) {
               if (audit.data.hasOwnProperty(key)) {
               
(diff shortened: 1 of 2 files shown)


suufi:
Could you elaborate on the necessity of this PR and how it solves multiple log occurrences?

conman0552:
This is a necessity as when listening through onAuditLog or onShout there are instances in which multiple events fire for the same thing. This appears to be something with the roblox API, and is more prevalent with onAuditLog. Specifically with onAuditLog, multiple audits appear but are identical, with the only thing being different about them is the milliseconds in the created value. To combat this, I've shaved off milliseconds when comparing the most recent Audit Log gets. So far I've not ran into any instances of duplicate logging anymore.

PR #162, opened 2020-07-17
Use latest endpoint to fetch X-CSRF-Token

policetonyR:


RGMW:
//api.roblox.com/sign-out/v1 => //auth.roblox.com/v1/logout

PR #166, opened 2020-07-20
JSON.parse(...).filter() is not a function fix

tobiO0310:
Add a function to the Object class with the name 'filter', that filters the object.
JSON.parse returns an Object class and since the Object class doesn't have a .filter() function, we create one that uses Object.keys and Array.reduce.

Feel free to improve this code, I am not really good at coding, but please don't trash talk. 😃

### DIFF
--- a/lib/group/getRankNameInGroup.js
+++ b/lib/group/getRankNameInGroup.js
@@ -5,6 +5,12 @@ var cache = require('../cache')
 // Args
 exports.required = ['group', 'userId']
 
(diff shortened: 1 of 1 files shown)


suufi:
https://api.roblox.com/users/55549140/groups returns an array. This is parsed from JSON form basically into an array you can work with in JavaScript. I don't understand why you're doing this considering an array is an `"object"` and not an `"Object"`.

PR #171, opened 2020-07-29
Add "setCookie" and "refreshCookie" declarations.

o-y:
Added two missing methods (user/setCookie and util/refreshCookie added in v4.6.0) to the typings file.
- The logic shown with setCookie describes a pattern in which the return types are conditional based on the parameters. By default a `Promise<LoggedInUserData>` is returned, however if the optional parameter `shouldValidate` is set to false then TypeScript infers the method response type will be a boolean.

| Arguments | Inferred type |
| ------------- | ------------- |
| `setCookie("*")`  | `Promise<LoggedInUserData>`  |
| `setCookie("*", true)`  | `Promise<LoggedInUserData>`  |
| `setCookie("*", false)`  | `boolean`  |

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for noblox.js@4.5.2
+// Type definitions for noblox.js@4.6.0
 // Authored by Gamenew09 w/ changes by suufi
 
 declare module "noblox.js" {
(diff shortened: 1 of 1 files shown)


Neztore:
Hi this is probably an issue of me badly naming the parameter.
As you can see here: https://github.com/suufi/noblox.js/blob/master/lib/user/setCookie.js
When validate is set to `false`, the return value is boolean.
When it is `true`, it is `LoggedInUserData`.

This is, I think, the inverse of the types you've defined?

o-y:
Ah, well spotted! My mistake, the typedefs were incorrect. I've just updated it to represent the behaviour defined in setCookie and updated the desc so that it correlates with the actual logic the the typedefs present.

Neztore:
Great, thanks for this!

PR #178, opened 2020-08-06
Implement getPlayersThumbnail() (Better than OptimisticSides' credits to him though for the idea)

Chosey98:
I have made a rework for how it works it should have all the typings and fully compatible with the thumbnails API and most of it's errors are checked manually and errored so I think this is ready to be merged


Chosey98:
Credits to Optimistic side for the old typings and old getPlayerThumnail and Neztore for suggesting how it could be better. I already changed all that lol

lib/user/getPlayersThumbnail.js
```diff
@@ -0,0 +1,47 @@
+// Includes
+var http = require('../util/http.js').func
```
Neztore, line 2: Please use `const` for imports that don't change

lib/user/getPlayersThumbnail.js
```diff
@@ -0,0 +1,47 @@
...
+
+// Args
+exports.required = ['userIds', 'size']
+exports.optional = ['format', 'isCircular']
+
+// Define
+exports.func = function (args) {
+	if(!Array.isArray(args.userIds)){
```
Neztore, line 10: To make it easier (and to stop the many questions we'll no-doubt get) how would you feel about making `userIds: number | number[]`?
And then implementing a check, something like:
```js
if (!Array.isArray(args.userIds)) {
   if (typeof userIds === "number") {
      userIds = [userIds]
   } else {
     //error
    }
}
```

lib/user/getPlayersThumbnail.js
Chosey98: Alright

lib/user/getPlayersThumbnail.js
Chosey98: Yeah I actually thought of that at first but then got confused while making the typings and totally forgot to create it

Chosey98:
Changes done @Neztore

PR #188, opened 2020-09-26
Fix configuring assets and buying them.

efewqefwfd:
Update the endpoints as the old ones are deleted. Fixed with Standard JS.

### DIFF
--- a/lib/asset/configureItem.js
+++ b/lib/asset/configureItem.js
@@ -1,39 +1,49 @@
 // Includes
-const generalRequest = require('../util/generalRequest.js').func
+const http = require('../util/http.js').func
+const getGeneralToken = require('../util/getGeneralToken.js').func
 
 // Args
 exports.required = ['id', 'name', 'description']
 exports.optional = ['enableComments', 'sellForRobux', 'genreSelection', 'jar']
 
 // Define
-function configure (jar, id, name, description, enableComments, sellForRobux, genreSelection) {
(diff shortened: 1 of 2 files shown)


lib/asset/configureItem.js
```diff
@@ -1,39 +1,49 @@
...
+      jar: jar,
+      headers: {
+        'X-CSRF-TOKEN': token
+      },
+      json: {
+        name: name,
+        description: description,
+        enableComments: enableComments,
+        genres: genreSelection || ['All'],
+        isCopyingAllowed: sellForRobux
```
Neztore, line 24: I notice this is a boolean in the typings, but the typings says:
```if `sellForRobux` is set it will be put on sale for that amount of robux.```

Your code seems to work with the fact it is a boolean, however - If sellForRobux is true your code will set copyingAllowed to true and there is no mention of a cost. Is this intended behaviour?

efewqefwfd: This value is intended for models and other items that cannot be sold for robux. I can add another parameter that will allow for the setting of price on the other endpoint.

efewqefwfd: I have added the ability to sell products for robux or set them as on sale.

Neztore:
Looks good. Thanks!

----
Noblox.js is participating in [Hacktoberfest](https://hacktoberfest.digitalocean.com/). Join it and this PR will count towards your free T-shirt!

PR #190, opened 2020-10-02
Add pastUsernames to getPlayerInfo

policetonyR:
I have tested this on my side and I got it to work. I just thought this would be a cool idea for people getting player info


Neztore:
Thanks for your PR - I've adjusted it slightly to use camelCase and return an array instead of a flat string.

Looks good, though. Cheers.

PR #193, opened 2020-10-06
Fix circular dependency error

nearbyplayer:
This appears to fix the circular dependency error when utilizing Node.js v14.


Neztore:
Hi, thanks for the PR!
I've just run this on v14 and it does not fix the issue.

PR #194, opened 2020-10-07
Update & Add errors/endpoints

efewqefwfd:
Add more descriptive errors and update endpoints to their newer counterparts.

### DIFF
--- a/lib/user/getCollectibles.js
+++ b/lib/user/getCollectibles.js
@@ -1,55 +1,17 @@
 // Includes
-const http = require('../util/http.js').func
+const getPageResults = require('../util/getPageResults.js').func
 
 // Args
 exports.required = ['userId']
-exports.optional = ['assetType', 'sortOrder', 'limit']
+exports.optional = ['assetType', 'sortOrder', 'limit', 'jar']
 
 // Define
-function getCollectibles (userId, assetType, sortOrder, limit, pageCursor, currentCollectibles) {
-  return new Promise((resolve, reject) => {
-    const allowedLimits = [10, 25, 50, 100]
-
-    const httpOpt = {
-      url: `//inventory.roblox.com/v1/users/${userId}/assets/collectibles`,
-      options: {
-        qs: {
-          assetType: assetType,
-          sortOrder: sortOrder || 'Asc',
-          limit: limit <= 100 ? allowedLimits.reduce((prev, curr) => Math.abs(curr - limit) < Math.abs(prev - limit) ? curr : prev) : 100,
-          cursor: pageCursor || ''
-        },
-        method: 'GET',
-        resolveWithFullResponse: true,
-        json: true
-      }
-    }
-    return http(httpOpt).then((res) => {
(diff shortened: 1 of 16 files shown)


Neztore (changes requested):
This looks good! I might release a -pre version with some of the new features to allow for thorough testing.

Could you please add typings for the new functions you've added? Thanks!

typings/index.d.ts
```diff
@@ -864,6 +867,46 @@ declare module "noblox.js" {
...
+    interface Collectibles {
+        collectibles: CollectibleEntry[];
+    }
+    //
+
+    interface InventoryEntry {
+        assetId: number;
+        name: string;
+        assetType: string;
+        created: string;
```
Neztore, line 899: Could we convert this to a Date?
Looks to be a standard ISO 8601 date string, so new Date(...) should be sufficient.

Neztore: Same goes for all other updated/created fields

typings/index.d.ts
```diff
@@ -864,6 +867,46 @@ declare module "noblox.js" {
...
+        username: string;
+        buildersClubMembershipType: number;
+    }
+    
+    interface CollectibleEntry {
+        userAssetId: number;
+        serialNumber?: number;
+        assetId: number;
+        name: string;
+        recentAveragePrice: string;
```
Neztore, line 884: Recent average price is a number, not a string.
According to test with my user id, `66592931`

lib/user/getCollectibles.js
```diff
@@ -0,0 +1,55 @@
...
+  return new Promise((resolve, reject) => {
+    const allowedLimits = [10, 25, 50, 100]
+
+    const httpOpt = {
+      url: `//inventory.roblox.com/v1/users/${userId}/assets/collectibles`,
+      options: {
+        qs: {
+          assetType: assetType,
+          sortOrder: sortOrder || 'Asc',
+          limit: limit <= 100 ? allowedLimits.reduce((prev, curr) => Math.abs(curr - limit) < Math.abs(prev - limit) ? curr : prev) : 100,
```
Neztore, line 19: What is going on here? Could you leave a comment explaining it?
Not just for my benefit, just for general maintainability.

Neztore (changes requested):
Thanks for a very high quality PR! These improvements are great.
Just a few little things to make this perfect:
- Typings mistakes
- Parsing dates. This makes it easier for the end user

Also,
Not a change required per-se, more of a future enhancement,  but is the cursor-handling behaviour consistent with other methods? Is this something that could be improved **in future**?
Perhaps the cursor handling logic could be moved to a separate file and shared between multiple methods that also make use of cursors. This would also ideally allow for streams/users customising how many pages they actually want to access.

Thanks for your work! :1st_place_medal:

Neztore (approved):
Looks good to me!
I'll merge this and put it out as -pre or -alpha at some point.

I haven't tested this extensively but I presume you have.

PR #197, opened 2020-10-09
Add error checking for getRank and getRankName

policetonyR:
I added a check to see if there was an error. Roblox only returns an error if the userid is fake/terminated i think. I got this to work to still work normally if userid/groupid was real.


Neztore:
Hi, thanks for your commits!
#194 provides similar functionality, however it is much more comprehensive and is, at least in my opinion, a better solution.

Please feel free to submit a PR with different fixes - there's still plenty to be fixed for hacktoberfest :)

Issue #192, opened 2020-10-06
Info: Hacktoberfest


Neztore:
## Noblox.js is participating in [Hacktoberfest](https://hacktoberfest.digitalocean.com/).

**Sign up on that website and submit 4, valid helpful pull requests (which are accepted) to participating open source repositories, and you can win a T-shirt!**

If you want to see if a repo is participating, look for the "hacktoberfest" topic.

### What sort of stuff needs done?
- Documentation (Make it automatically generated, update it - Could use TS Typings/JSDOC?)
    * Linked issue: https://github.com/suufi/noblox.js/issues/9
**Documentation is a priority**!

- Investigate and resolve issues with circular dependencies in v14.
   * This is a very important fix that is needed as soon as possible
   * this is quite an involved/difficult task. I've started some work towards it and would be happy to work with you on fixing it.
       see #195 for more info!

- Add new features
   * There are a lot of missing features that could be added
   * You could prepare a manual/automated coverage report on how much of the API noblox covers.

- Improve errors and make them more consistent
    ~~* Linked: https://github.com/suufi/noblox.js/issues/181~~
        * This is something that could be improved throughout the library. The linked issue has a PRed fix, however.

~~- Investigate getGroups issues
    * Linked: https://github.com/suufi/noblox.js/issues/187~~

### Big improvements you could make
- Work on moving away from `request`. Preferably to `node-fetch` or something similar! (v5 has some progress - speak to Suufi/Qxest)
- Work on v5 (see above). Should try to use more a OOP-style
- Improve events
    * Events have some issues with emitting connection reset errors. Could these be prevented?
    * Allow for event delay to be customisable on every function call (and try to move away from using a json file)

Please do not submit invalid requests or ones that fix tiny issues like spacing. Unless it's actually a problem!
Please make sure all code you submit complies with StandardJS. Check this with npm test before PRing.

If you'd like to tackle any of these please let me know and I'll assign the relevant issue to you.
I'm happy to help if you come across any problems while working on the library. Cheers!

Sign up [here](https://hacktoberfest.digitalocean.com/).

Neztore:
Thanks to those who took part!

PR #198, opened 2020-10-11
Handle wrong userId, and return error

xIgreq:


Neztore:
Hi, thanks for your commits!
#194 provides similar functionality, however it is much more comprehensive and is, at least in my opinion, a better solution.

Please feel free to submit a PR with different fixes - there's still plenty to be fixed for hacktoberfest :)
See #192 for an up to date list of what needs done!

PR #199, opened 2020-10-12
gh-pages update

conman0552:
Started adding side tabs for events and updated the top to mention stuff about refreshing cookies. Planning to go back through and make additional changes to that, as well as actually build the pages for the side tabs. Have not pushed them in this PR because someone else might want to also help make those pages.


Neztore (approved):
looks good to me

PR #201, opened 2020-10-12
Fix GetStatus typo

jayden8923:
The GetStatus documentation reads as:

```
Get Status

This function returns the blurb of a user. The blurb is the equivalent of the user's description on Roblox, not their status message.
```
It should say:

```
Get Status

This function returns the status of a user. The blurb is the equivalent of the user's description on Roblox, not their status message.
```

### DIFF
--- a/index.html
+++ b/index.html
@@ -1094,7 +1094,7 @@ Note: By default, setCookie will validate the cookie you provide by making a HTT
 </tr>
 </tbody></table>
 <h2 id='get-status'>Get Status</h2>
-<p>This function returns the status of a user. The blurb is the equivalent of the user&#39;s description on Roblox, not their status message.</p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1094,7 +1094,7 @@ <h2 id='get-presences'>Get Presences</h2>
 </tr>
 </tbody></table>
 <h2 id='get-status'>Get Status</h2>
-<p>This function returns the blurb of a user. The blurb is the equivalent of the user&#39;s description on Roblox, not their status message.</p>
+<p>This function returns the status of a user. The blurb is the equivalent of the user&#39;s description on Roblox, not their status message.</p>
```
Neztore, line 1097: Could you please also replace the other instance of "blurb" with status?
Are there any other improvements you could make, to make this PR even better?

PR #202, opened 2020-10-13
Add getUserBadgeTimestamp

policetonyR:
This feature uses the https://badges.roblox.com/docs#!/BadgeAwards/get_v1_users_userId_badges_awarded_dates api

It allows you to see the timestamp of when a user is awarded a badge.

### DIFF
--- a/lib/badges/getUserBadgeTimestamp.js
+++ b/lib/badges/getUserBadgeTimestamp.js
@@ -1,0 +1,40 @@
+// Includes
+const http = require('../util/http').func
+
+// Args
+exports.required = ['userId','badgeId']
+
+// Define
+const getBadgeTimestamp = (userId, badgeId) => {
+  return new Promise((resolve, reject) => {
+  const httpOpt = {
+    url: `https://badges.roblox.com/v1/users/${userId}/badges/awarded-dates?badgeIds=${badgeId.join(',')}`,
+    options: {
+      method: 'GET',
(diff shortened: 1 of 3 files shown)


Neztore:
What do you think about renaming this method to getBadgeAwardedTimestmap or similar, like getBadgeAwarded or even getBadgeAwarded date?

Just seems more clear, in my opinion. Open to whatever, though!

policetonyR:
I only named it this because I did not know what else to name it. You could maybe do getBadgeAwardedTimestamps since it takes an array like getPresences.

lib/badges/getUserBadgeTimestamp.js
```diff
@@ -0,0 +1,40 @@
+// Includes
+const http = require('../util/http').func
+
+// Args
+exports.required = ['userId','badgeId']
+
+// Define
+const getBadgeTimestamp = (userId, badgeId) => {
```
Neztore, line 8: Rename this method (The file and the function itself plus the typings) to getAwardedTimestamps.

I do not think there are any other assets that can be "awarded" and this avoids plurality surrounding "badge(s)" in the method name.

Neztore (changes requested):

lib/badges/getUserBadgeTimestamp.js
policetonyR: Added

Neztore (approved):

PR #205, opened 2020-10-18
Update examples

### LINKED ISSUE #191: Suggestion, Examples.
I would like to see on web or on git hub some examples on what how to use so anyone who has little - no experience can get used to promises and how they work. Like for example for setRank how to get a role name from it. (roblox = noblox.js)
```js
roblox.setRank(groupId, userId, rankIdentifier)
.then(function(newRole){
message.channel.send(`User's rank has been changed to ${newRole.name}!`)
```

That was just an example what you could add. I just thing it could help not only me but many other people. I'm getting issue when I'm trying to get all roblox group rank's it always return's undefined or some array that I don't know how to resolve. I know it might seem simple to some nerds out there but for new people to arrays and promises its kinda rough. And the page that is about promises I can't really understand because I'm not native in English. I'm trying my best to understand it, but sometimes small help like some examples on how to use something. It could help a lot I think. I don't know how to resolve the getRoles array and that's all I can't for now. I'd appreciate if someone could help me with the code. I have been for the past...... 3? hours trying to resolve the getRoles array because I couldn't get to know why it shows undefined or an array of thing's that I can't use since if I try to use it, it just says that it isn't defined.

GeneralScripter:
Previous examples were extremely outdated and broken, these provide similar (and in some cases extended) functionality.
They are all fairly common uses of the module so it's a good resource for those who want to learn from examples and those who just want to copy and paste.

Resolves #191


GeneralScripter:
Resolves #191

PR #213, opened 2020-10-29
Update developer product endpoints and improve on obscure errors

GeneralScripter:
Endpoints were outdated and likely to be removed at some point, they are now updated to use the supported API.
Old implementation gave a generic error message due to the old endpoints not providing any information. This is no longer the case, the functions return any errors thrown by the API.
Typings was actually inconsistent with the arguments the functions were expecting. This has been changed so the functions are now consistent with typings.
Added iconImageAssetId argument.

### DIFF
--- a/lib/game/addDeveloperProduct.js
+++ b/lib/game/addDeveloperProduct.js
@@ -2,7 +2,7 @@ const http = require('../util/http.js').func
 const getGeneralToken = require('../util/getGeneralToken.js').func
 
 exports.required = ['universeId', 'name', 'priceInRobux']
-exports.optional = ['prodDescription', 'iconImageAssetId', 'jar']
+exports.optional = ['description', 'iconImageAssetId', 'jar']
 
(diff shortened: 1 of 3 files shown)


suufi:
The reason I'm not going to merge this yet is that it doesn't behave predictably. Please look at [this](https://devforum.roblox.com/t/inconsistent-behavior-on-the-develop-apis-developer-product-endpoints/775969) for more information.

GeneralScripter:
> The reason I'm not going to merge this yet is that it doesn't behave predictably. Please look at [this](https://devforum.roblox.com/t/inconsistent-behavior-on-the-develop-apis-developer-product-endpoints/775969) for more information.

Wow, I completely understand. I didn't test for accuracy of the returned values because I had assumed it would be correct. As for the update endpoint, I didn't realise it behaved that way. I'll be sure to update this PR as necessary if and when they resolve the issues with those new endpoints.

PR #214, opened 2020-11-01
Add documentation

efewqefwfd:
Added documentation across the entire library & added missing TypeScript typings. Also added tutorials that still need to be revised/added to.


docs/tutorials/Authentication.md
```diff
@@ -0,0 +1,51 @@
+### What is Authentication on Roblox?
+Authentication is pretty simple, it's how Roblox knows who you are and what account you're using. Since Roblox uses FunCaptcha and have no official bot system, developers are forced to use account [**cookies.**](https://support.mozilla.org/en-US/kb/cookies-information-websites-store-on-your-computer) This will allow Roblox to easily identify your bot's account and execute actions.
```
Neztore, line 2: Would be nice to have some explicit language that says that we use an **authentication token** which happens to be stored within cookies. And it links to an authentication session.

lib/asset/uploadModel.js
```diff
@@ -5,6 +5,32 @@ const http = require('../util/http.js').func
 exports.required = ['data']
 exports.optional = ['itemOptions', 'asset', 'jar']
 
+// Docs
+/**
+ * Upload a model.
+ * @category Assets
+ * @alias uploadModel
+ * @param {Buffer} data - The model data.
```
Neztore, line 13: Buffer does not match the `uploadItem` method nor the [Typescript typings](https://github.com/suufi/noblox.js/blob/master/typings/index.d.ts#L984) - is it a buffer or a Stream?

lib/avatar/getCurrentAvatar.js
```diff
@@ -1,7 +1,20 @@
+/* @module Avatar */
```
Neztore, line 1: Is this meant to be here? I note none of the other avatar-related files have a `@module` decorator.

lib/badges/getAwardedTimestamps.js
```diff
@@ -4,6 +4,18 @@ const http = require('../util/http').func
 // Args
 exports.required = ['userId', 'badgeId']
 
+// Docs
+/**
+ * Get the time the badge was awarded to a user.
+ * @category Badges
+ * @alias getAwardedTimestamps
+ * @param {number} userId - The userId of the user.
+ * @param {Array} badgeIds - The badge's id.
```
Neztore, line 13: Pluralisation - The badge ids

Neztore (changes requested):
Preliminary review - I haven't reviewed all the files but I'm going to check over the typings before finishing this.
I may fix some of these issues, or if you could that'd be great. If you object to any, lmk.

suufi (approved):
I've corrected any mistakes I saw and revised it so that it targets the docs/ folder for deployment.

Neztore (approved):

Issue #195, opened 2020-10-07
[BUG] Noblox.js causes circular dependency warning on node v14


Neztore:
**Describe the bug**
The following is outputted:
```(node:41901) Warning: Accessing non-existent property 'func' of module exports inside circular dependency
    at emitCircularRequireWarning (internal/modules/cjs/loader.js:672:11)
    at Object.get (internal/modules/cjs/loader.js:686:5)
    at Object.<anonymous> (/home/josh/Desktop/test/efewqefwfd/lib/index.js:23:16)
    at Module._compile (internal/modules/cjs/loader.js:1085:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1114:10)
    at Module.load (internal/modules/cjs/loader.js:950:32)
    at Function.Module._load (internal/modules/cjs/loader.js:791:14)
    at Module.require (internal/modules/cjs/loader.js:974:19)
    at require (internal/modules/cjs/helpers.js:88:18)
    at Object.<anonymous> (/home/josh/Desktop/test/index.js:1:11)
    at Module._compile (internal/modules/cjs/loader.js:1085:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1114:10)
    at Module.load (internal/modules/cjs/loader.js:950:32)
    at Function.Module._load (internal/modules/cjs/loader.js:791:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:72:12)
    at internal/main/run_main_module.js:17:47```

**To Reproduce**
Steps to reproduce the behavior:
1. Run on node v14.
Sample test code:
```js
const n = require("./noblox.js");
async function test () {
    console.log(await n.getPlayerInfo(66592931));

}
test();

```

**Expected behavior**
Runs fine.

## Debugging done so far
I have set NODE_DEBUG=module and have got the following:
```MODULE 41901: RELATIVE: requested: ../../settings.json from parent.id /home/josh/Desktop/test/efewqefwfd/lib/util/threaded.js
MODULE 41901: looking for ["/home/josh/Desktop/test/efewqefwfd/lib/util"]
MODULE 41901: Module._load REQUEST ./options.js parent: /home/josh/Desktop/test/efewqefwfd/lib/index.js
MODULE 41901: RELATIVE: requested: ./options.js from parent.id /home/josh/Desktop/test/efewqefwfd/lib/index.js
[...]

PR #218, opened 2020-11-08
Update index.js to fix accessing of non-existent property

unix-system:
Fix bad reference that causes warnings when used with TypeScript

This isn't really important, since JS *manages*, but it's nice and will get rid of annoying warnings from my runtime.

The warning was: 

``(node:1572) Warning: Accessing non-existent property 'func' of module exports inside circular dependency``

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -19,7 +19,7 @@ search(__dirname)
 
 for (const name in rbx) {
   const exporter = rbx[name]
-  if (exporter.hasOwnProperty("func")) {
+  if (Object.prototype.hasOwnProperty.call(exporter, 'func')) {
     module.exports[name] = rbx.wrap.wrapExport(exporter.func, exporter.required || [], exporter.optional || [])
   } else {
     module.exports[name] = rbx[name]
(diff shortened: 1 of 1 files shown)


Neztore:
This isn't an issue with Typescript. it's an issue with node v14 onwards.
A fix was submitted previously which did exactly what you've done. I tested it, and it didn't fix the issue for me.
However I've now tested your code and it appears to be working.
Closes #195

Neztore:
Hi, this fails the standardJS test.

unix-system:
Weird - I guess there is some standard misfitting with node v15, TypeScript and StandardJS...

suufi:
They used to be included but thank you for reminding me to migrate Travis CI from the legacy service. All PRs are now run against `yarn test` which should hopefully prevent future "Adhere to standard commits."

lib/index.js
```diff
@@ -19,7 +19,7 @@ search(__dirname)
 
 for (const name in rbx) {
   const exporter = rbx[name]
-  if (exporter.func) {
+  if (exporter.hasOwnProperty("func")) {
```
suufi, line 22: See [standard rules](https://eslint.org/docs/rules/no-prototype-builtins)

suufi (changes requested):

suufi (approved):

PR #219, opened 2020-11-08
Add missing typing "getGroups"

unix-system:
Missing typings for "getGroups"

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1355,6 +1355,12 @@ declare module "noblox.js" {
     function getPlayerBadges(userId: number, limit?: Limit, cursor?: string, sortOrder?: SortOrder): Promise<PlayerBadges>
 
     /**
+     * Gets aa users groups
+     */
+    function getGroups(userId: number): Promise<Group[]>;
+
+
+    /**
      * Gets a brief overview of a user.
      */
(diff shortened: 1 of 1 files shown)


typings/index.d.ts
```diff
@@ -1354,6 +1354,12 @@ declare module "noblox.js" {
     */
     function getPlayerBadges(userId: number, limit?: Limit, cursor?: string, sortOrder?: SortOrder): Promise<PlayerBadges>
 
+    /**
+     * Gets aa users groups
```
suufi, line 1358: Typo, "Gets a user's groups"

typings/index.d.ts
```diff
@@ -1354,6 +1354,12 @@ declare module "noblox.js" {
...
+     */
+    function getGroups(userId: number): Promise<Group[]>;
```
suufi, line 1360: It actually doesn't return the GroupInterface. It's rather similar but it has some differences, see below:
image. You might want to add a new interface for this.

suufi (changes requested):

typings/index.d.ts
unix-system: Yep - I realised this a bit late on. I've done some modifications locally and I'm testing them, then I'll commit n' push. Stand-by!

PR #230, opened 2020-11-29
Add index typing for IGroupPartial

unix-system:
Fixes #219

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1193,6 +1193,9 @@ declare module "noblox.js" {
      */
     function getGroup(groupId: number): Promise<Group>;
 
+    /**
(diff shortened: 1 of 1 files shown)


typings/index.d.ts
```diff
@@ -1182,6 +1193,8 @@ declare module "noblox.js" {
      */
     function getGroup(groupId: number): Promise<Group>;
 
+    function getGroups(userId: number): Promise<IGroupPartial[]>
```
suufi, line 1196: Even though it's obvious, it would be helpful to add a description of what the function does in a comment above.

suufi (changes requested):

suufi (approved):

PR #235, opened 2020-12-09
Add more features to getplayerinfo

policetonyR:
Adds friendcount, followingcount, and followercount to the function.


Neztore:
Looks good to me!

PR #250, opened 2021-01-05
Added configureGamePass() method.

alanbixby:
I added a `configureGamePass()` method since the existing `configureItem()` is incompatible with GamePassIDs, as GamePassIDs can overlap AssetIDs. 

I tried to maintain parity with the structure used by `configureItem()` as much as possible, but the `www.roblox.com/game-pass/update/` endpoint forces weird behavior as it is seemingly designed for the web browser rather than a REST API.

My method takes the parameters:
```js
configureGamePass(gamepassId: number, name: string, description?: string, price?: number | boolean, jar?: CookieJar)
```

Roblox breaks the dashboard up into `name`/`description` and `price`/`isForSale`; you cannot change both groups with a singular API call, ergo the WET code that makes up to two API calls; unnecessary calls are automatically dropped. Passing `0`, `false` or a negative value to `price` will make the game pass go off sale.

I have noted the quirk in `index.d.ts` that passing only a `name` updates both the `name` and `description`; thus purging the old description.

---

Since this uses the `www.` subdomain, errors display via an HTML page rather than a JSON response. I think I have any type errors covered and have left an intentionally vague error message for bad GamePassIDs throw call.

**TODO:** Currently this method doesn't support changing the icon. I couldn't justify the time investment for the niche within the niche use-case as this method already supports my needs and I'd have to fuss with ROBLOX moderation while testing.

### DIFF
--- a/lib/game/configureGamePass.js
+++ b/lib/game/configureGamePass.js
@@ -2,12 +2,38 @@
 const http = require('noblox.js/lib/util/http').func
 const getGeneralToken = require('noblox.js/lib/util/getGeneralToken').func
 
+// Docs
+/**
+ * Modifies an existing game pass.
+ * @category Game
+ * @alias configureGamePass
+ * @param {number} gamePassId - The id of the game pass.
+ * @param {string} name - The name of the game pass; skips name, description, and icon if set to "".
+ * @param {string=} description - The description of the game pass; description is updated when name is modified.
+ * @param {number|boolean=} price - The price of the game pass in Robux; sets to 'Off Sale' if 0, false, or a negative value; skips if true.
+ * @param {ReadStream=} icon - The read stream for the game pass icon being uploaded; .png, .jpg, .gif
+ * @returns {Promise<GamePassResponse>}
+ * @example const noblox = require("noblox.js")
+ * const fs = require("fs")
+ * // Login using your cookie
+ * noblox.configureGamePass(12345678, "Game Pass Title", "Game Pass Description", 1234, fs.createReadStream("./Image.png"))
+**/
+
 // Args
-exports.required = ['id', 'name']
-exports.optional = ['description', 'price', 'jar']
+exports.required = ['gamePassId', 'name']
+exports.optional = ['description', 'price', 'icon', 'jar']
 
 // Define
-function configure(jar, token, id, name, description, price) {
+function configure (jar, token, gamePassId, name, description, price, icon) {
+  const file = icon
+    ? {
+        value: icon,
+        options: {
+          filename: 'icon.png',
+          contentType: 'image/png'
+        }
+      }
+    : undefined
   const httpOpt = {
     url: '//www.roblox.com/game-pass/update',
     options: {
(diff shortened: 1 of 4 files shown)


lib/game/configureGamePass.js
```diff
@@ -0,0 +1,100 @@
+// Includes
+const http = require('noblox.js/lib/util/http').func
+const getGeneralToken = require('noblox.js/lib/util/getGeneralToken').func
+
+// Args
```
suufi, line 5: Document arguments and return type, see other files that have jsdoc documentation.

typings/index.d.ts
```diff
@@ -1200,6 +1202,14 @@ declare module "noblox.js" {
 
     function updateDeveloperProduct(universeId: number, productId: number, name: string, priceInRobux: number, description?: string, jar?: CookieJar): Promise<DeveloperProductUpdateResult>;
 
+    /**
+     * Configures a gamepass with the id `gamepassId` to have a `name` and `description`, and sets the `price` in Robux. If both `name` and `description` are empty, only `price` is changed. Setting `price` to false, 0, or a negative value will place the gamepass off-sale.
+     * Returns `gamepassId` with the changed attributes.
+     * 
+     * NOTE: Updating `name` will affect `description`: you must 'refresh' `description` with each `name` update, or `description` will be cleared.
+     */
+    function configureGamePass(gamepassId: number, name: string, description?: string, price?: number | boolean, jar?: CookieJar): Promise<GamepassInfo>;
```
suufi, line 1211: GamepassInfo isn't a valid interface. You'll want to include it towards the top of the file and in the other typings file.

lib/game/configureGamePass.js
```diff
@@ -0,0 +1,100 @@
...
+exports.required = ['id', 'name']
```
suufi, line 6: 'id' should be 'gamepassId' (consistency)

lib/game/configureGamePass.js
```diff
@@ -0,0 +1,100 @@
...
+exports.optional = ['description', 'price', 'jar']
```
suufi, line 7: You can include image changes: https://github.com/suufi/noblox.js/blob/master/lib/asset/uploadItem.js
I don't think it's that different from uploadItem

suufi (changes requested):
Thank you for the PR. Could you please review the following changes?

suufi:
Your commit also doesn't pass standard rules. See:
image

Please run `npm run lint` or `yarn lint` to make sure your code style is consistent with the rest of the library.

lib/game/configureGamePass.js
alanbixby: Added. Opted for `gamePassId` in favor of `gamepassId` to be consistent with Roblox's "game-pass" naming scheme.

typings/index.d.ts
alanbixby: Added as `GamePassResponse` interface; seems I missed this one when manually transposing my edits from my working bot directory to the cloned repo- just first pull request things I guess 🙄

alanbixby:
I believe I have applied the requested changes, images are now supported and error handling is bit more specific. Apologies for the linting and interface issues, as it seems I had a transposing error moving from my node_module edits to the cloned repo.

suufi (approved):

PR #254, opened 2021-01-08
Fix getPlayerInfo

popeeyy:
Fixed getPlayerInfo, updated typings, and made it consistent with the rest of the library.

### DIFF
--- a/lib/user/getPlayerInfo.js
+++ b/lib/user/getPlayerInfo.js
@@ -40,11 +40,11 @@ function getPlayerInfo (userId) {
     Promise.all(requests).then(function (responses) {
       const profileResponse = responses[0]
       const usersResponse = responses[1]
+      const userBody = usersResponse ? usersResponse.body : {}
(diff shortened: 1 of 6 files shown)


lib/user/getPlayerInfo.js
```diff
@@ -18,33 +17,48 @@ exports.required = ['userId']
...
+          json: true
+        }
+      })
+    ]
+
+    Promise.all(requests).then(function (responses) {
+      const profileResponse = responses[0]
+      const usersResponse = responses[1]
+
+      if (profileResponse.statusCode === 200 && usersResponse.statusCode === 200) {
```
alanbixby, line 44: This can't resolve banned users.

`profileResponse` has status code 302 when a banned user is called, thus causing the method to throw error 'User does not exist'.

Use the Roblox ID, [`8166491`](https://users.roblox.com/v1/users/8166491) to test this.

PR #256, opened 2021-01-14
Update getOwnership() to support game passes, badges, and bundles.

alanbixby:
Currently `getOwnership()` only supports assetIDs; I changed the endpoint being used so it now supports an optional parameter for the alternative item types: `GamePass`, `Badge`, and `Bundle`.

All changes should maintain backwards compatibility. I also added three unit tests using [`Qxest`](https://www.roblox.com/users/55549140/profile) which covers the three additional ID types.

---

Minor documentation hotfixes to `getMessages()` and `configureGamePass()` for parity are included as well.


suufi (approved):
Perfect, thank you!

PR #273, opened 2021-02-08
Added getPremium() function

Khristian-Sutherland:
Added a new function `getPremium()` that takes a user id, and returns `true` or `false` depending on whether the user with the given userId has premium or not. This uses the following endpoint:
`https://premiumfeatures.roblox.com/v1/users/${userId}/validate-membership`
This endpoint requires the bot to be logged in.
Feel free to user a better name than `getPremium`.

`// login with cookie`
`const hasPremium = await noblox.getPremium(123456)`

### DIFF
--- a/lib/user/getPremium.js
+++ b/lib/user/getPremium.js
@@ -18,34 +18,34 @@ exports.optional = ['jar']
  **/
 
 // Define
-function getPremium(jar, userId) {
-    return new Promise((resolve, reject) => {
-        const httpOpt = {
-            url: `https://premiumfeatures.roblox.com/v1/users/${userId}/validate-membership`,
-            options: {
-                method: 'GET',
-                jar: jar,
-                resolveWithFullResponse: true
-            }
-        }
-        return http(httpOpt)
-            .then(function(res) {
-                if (res.statusCode === 200) {
-                    resolve(res.body == 'true')
-                } else {
-                    const body = JSON.parse(res.body) || {}
-                    if (body.errors && body.errors.length > 0) {
-                        const errors = body.errors.map((e) => {
-                            return e.message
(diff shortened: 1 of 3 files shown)


lib/user/getPremium.js
```diff
@@ -0,0 +1,51 @@
...
+exports.required = ['userId']
+exports.optional = ['jar']
+
+// Docs
+/**
+ * Gets whether or not a user has premium.
+ * @category User
+ * @alias getPremium
+ * @param {number} userId - The id of the user.
+ * @returns {Promise<void>}
```
alanbixby, line 14: Return type should be `Promise<boolean>`.

alanbixby, line 14: Return type should be `boolean` instead of `void`.
```suggestion
 * @returns {Promise<boolean>}
```

alanbixby (changes requested):
Small type error in JSDoc info of the main file, otherwise good first pull request. 😀

---

And as a comment, there was some talk in the Roblox API Discord of merging this into [`getPlayerInfo()`](https://github.com/suufi/noblox.js/blob/master/lib/user/getPlayerInfo.js), but I agree with the consensus that leaving this as an independent method for the time being is ideal due to it's need for authentication.

suufi:
Could you send screenshots of you using this function on premium/non-premium members?

Khristian-Sutherland:
> Could you send screenshots of you using this function on premium/non-premium members?

Thanks for asking, or I wouldn't have found an issue. It works fine now though.
https://gyazo.com/43d54e21deca72f6ff230622e55cdf51
https://gyazo.com/792382b00d28a94590cf81ce2cc1a1bc

PR #274, opened 2021-02-10
Upgrade getPlayerThumbnail() to support bust and headshot

alanbixby:
Currently the library has no method to retrieve player headshots. The endpoint used by `getPlayerThumbnail()` is nearly identical to the headshot endpoint, so I have added an optional fifth parameter to retrieve headshots. 

Minimal code clean up and increased specificity to typing are included. **All changes are backwards compatible with the old implementation.**


suufi (changes requested):
image

So there's an issue with the two endpoints. The headshot one has a different set of sizes. I think we would have to do something similar to lib/group/getLogo.js line 14 and have a mixed parameter (e.g. `{ThumbnailSizes|HeadshotSize}`).

alanbixby:
I rewrote `getPlayerThumbnail()` so now it supports `Body`, `Bust` and `Headshot` thumbnails. Tldr is changes add bust and make error handling easier for the end user.

---

A quirk about the endpoint is that uncached users (Roblox's cache) and banned users will return objects of status 'Pending' and 'Blocked' respectively and have `imageURL` set to `null`. As such, I've added a config to `settings.json`, and a folder in `img` to point the `null` thumbnails to `https://noblox.js.org/moderatedThumbnails/moderatedThumbnail_{size}.png`.



`Pending` responses tend to resolve themselves with a follow-up request, so `getPlayerThumbnail()` recursively calls itself up to `settings.thumbnail.maxRetries` (2) additional times to resolve pending thumbnails. If this fails, then it falls back to the `noblox.js.org` thumbnail link, or a URL that the user can specify in `settings.json`.

---

For the size parameter, users can provide a number or a string for the dimension. `150` <-> `"150x150"`; if no size is provided it defaults to the largest possible size, and if an invalid size is provided, the error contains the array of valid sizes.

Documentation in `index.d.ts` uses type definitions of `BodySize`, `BustSize`, and `HeadshotSize`, while the `JSDocs` documentation will only state 'number | string`. My reasoning is the type definitions don't translate well to the website's documentation, and instead we can provide a link to this table in the parameter description.

suufi (approved):

PR #300, opened 2021-03-25
Using a ternary operator when setting the primaryGroupId variable.

realCartar:
I've ran into a long list of issues with this function. If the user doesn't have a set primary group, it errors and throws a TypeError (not poggers). Switching to a simple ternary operator completely fixes this issue.

### DIFF
--- a/lib/user/getGroups.js
+++ b/lib/user/getGroups.js
@@ -48,7 +48,7 @@ function getGroups (userId) {
 
       const groupRoleData = responses[0].data
       if (groupRoleData) {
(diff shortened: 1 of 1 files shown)


alanbixby:
This issue should be resolved by #299 but hasn't been published to the npm package yet. 

Given that its a verified commit I assume they spotted the issue and made the PR in the browser instead of it actually needing to be fixed (again).

Neztore:
Thanks for the PR - it's already been fixed as described above.

PR #309, opened 2021-03-29
Fixed typings issue

LightYagami200:
Replaced `ThumbnailSizes` to `HeadshotSizes` for method `getPlayerThumbnail`

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1552,7 +1552,7 @@ declare module "noblox.js" {
     /**
      * Gets the thumbnail of an array of users.
      */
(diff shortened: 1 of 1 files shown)


alanbixby:
**Context:**
https://github.com/suufi/noblox.js/blob/3d275a5ca2584815d41d5b33c6bd30bb119956e6/typings/index.d.ts#L771-L774

https://github.com/suufi/noblox.js/blob/59449074a542b54776d854c1faa943569365c51f/typings/index.d.ts#L1555

alanbixby (approved):
Good catch; thank you for spotting the type error.

PR #311, opened 2021-04-01
Remove Role.ID in favour of role.id

Neztore:
Not sure why we convert role.id to role.ID? I imagine it was in favour of past consistency.
The typings show role has an `id` field, not `ID`. This is not the case.
This brings the library in-line with the typings, but retains backwards compatibility.


suufi (approved):

BananCZE:
Hello,
thank you for the commit! It works fine for me now, but I'll keep using directly Roblox's Api endpoint as I find it more performant and reliable.

Thank you!

PR #312, opened 2021-04-02
Added a new getGroupGames function

Vilksian:
I noticed there wasn't any convenient way to get the games owned by a group, so I made a function for it.

### DIFF
--- a/lib/group/getGroupGames.js
+++ b/lib/group/getGroupGames.js
@@ -4,7 +4,7 @@ const getPageResults = require('../util/getPageResults.js').func
 
 // Args
 exports.required = ['groupId']
-exports.optional = ['accessFilter', 'sortOrder', 'limit', 'cursor']
+exports.optional = ['accessFilter', 'sortOrder', 'limit']
 
 // Docs
 /**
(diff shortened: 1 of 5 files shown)


popeeyy:
getGroupGames should be using [this](https://github.com/suufi/noblox.js/blob/master/lib/util/getPageResults.js) function as it uses cursors. There is an example of this [here](https://github.com/suufi/noblox.js/blob/master/lib/user/getInventory.js).

Vilksian:
> getGroupGames should be using [this](https://github.com/suufi/noblox.js/blob/master/lib/util/getPageResults.js) function as it uses cursors. There is an example of this [here](https://github.com/suufi/noblox.js/blob/master/lib/user/getInventory.js).

Thanks. Changed the function to work with that.

Neztore (changes requested):
Please squash your commit, i.e. removing changes that are already on master.

Thanks!

Vilksian:
> Please squash your commit, i.e. removing changes that are already on master.
> 
> Thanks!

I'm not sure what you mean by that. What parts are already in the master branch?

suufi:
Closed by #345

PR #343, opened 2021-05-24
Update docs: mark authentication requirements, add assetType hyperlinks, deprecate refreshCookie, update getPlayerThumbnail bust sizes, update README.md

alanbixby:
I wanted to add a more visual way to tell if an endpoint required `setCookie()` before it could be used, previously you had to rely on the example snippets having `// Login using your cookie` which is rather inconsistent (some had it and didnt need it, others needing it but missing it), and comments are inherently low contrast.

As such, in addition to making the `// Login using your cookie` comments consistent, I added an emoji legend to the documentation descriptions:
🔐: indicates authentication is required to use the method
🔓 : indicates authentication _may be required_, such as `getInventory()`, or `getShout()`, where privacy settings may restrict access to certain users
✅ : indicates no authentication is required to use the method

~~**NOTE:** Currently this isn't labelled anywhere explicitly, but I opted to keep the README.md clean since it might cause confusion if someone read the GitHub page but not the docs.~~

Went ahead and redid the README while I was at it; I removed all the references to `rbx` and replaced it with `noblox` for more consistency.

---
**Other tid-bits:**
- I added hyperlinks to the [Roblox Wiki's `AssetTypes` page](https://developer.roblox.com/en-us/api-reference/enum/AssetType) for relevant methods such as `getInventory()`, in addition to a warning about badges being incompatible for the method (caused by Roblox API 500ing, and has been like that for months).
- I marked `refreshCookie()` as deprecated due to incognito cookies never expiring with a hyperlink to the authentication tutorital.
- Updated `getPlayerThumbnail()`'s bust sizes to reflect API changes (more options, bigger images)


alanbixby:
wiQPd2J

Neztore:
I'm not sure if we want to deprecate refreshCookie.
It's not required in most use-cases, but it's still a perfectly valid method.

alanbixby:
> I'm not sure if we want to deprecate refreshCookie.

I sympathize with the sentiment, but this change is limited to JSDocs only. The marking targets individuals following old tutorials that rely on `refreshCookie()` logic, which I've seen pop up a handful of times in the support server. 

Regardless, it's on a `www.` endpoint, and it isn't in the Jest suite, so we likely wouldn't even notice, let alone be able to fix it, if Roblox ever moved the functionality. ¯\\\_(ツ)\_/¯

alanbixby:
Im an idiot and mixed up my branches when updating #345.. ignore my small mess.

Neztore (approved):
Great work! I have a few concerns:
- Do we want to explicitly deprecate refreshCookie in the docs? I think you raised good points so I'd be open to it, as relatively few cases need it.
- Is it really necessary to rename all `rbx` to `noblox`? It's pretty harmless so I don't think this is a blocking thing, I just don't see the advantage - Noblox is used to interact with Roblox, Roblox is abbreviated to `rbx`.

That being said, I'd be happy for this to be merged

PR #344, opened 2021-05-25
Change all "target" userId parameters to "userId"

Zenthial:
Many of the group functions take either "target" or "userId" as their parameter name for userId. This creates confusion on the documentation. An example of this is getRankInGroup(group, userId), which has its userId parameter labeled "userId", while setRank(group, target, rank) has it's userId parameter labeled "target". This is an unnecessary dual labeling when it is the same parameter, and should be fixed.

### DIFF
--- a/examples/cleanPlayers.js
+++ b/examples/cleanPlayers.js
@@ -50,31 +50,31 @@ rbx.setCookie(cookie)
     }
 
     const groupRoles = await rbx.getRoles(getRolesOptions)
-    const targetRoles = []
+    const userIdRoles = []
 
     for (const role of groupRoles) {
       if (options.roleset.includes(role.id)) {
-        targetRoles.push(role)
+        userIdRoles.push(role)
       }
 
-      if (targetRoles.length === options.roleset.length) {
+      if (userIdRoles.length === options.roleset.length) {
         break
       }
     }
 
-    if (targetRoles.length === 0) {
+    if (userIdRoles.length === 0) {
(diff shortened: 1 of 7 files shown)


suufi (changes requested):
Not all "target" references in noblox are referring to userIds (e.g. cleanPlayers). I think this is mostly a continuation from roblox-js and a time where Roblox used `target`/`targetUserId` with regards to web APIs. Please double check your PR and provide screenshots that methods still work as expected. 

P.S.:  `({ group: group, userId: userId })` can turn into `({ group, userId })`

alanbixby:
Also be mindful this proposal would introduce a breaking change to anyone that uses object parameters.

Thar said, I'm in the process of rewriting `wrap.js` so it can give better errors and accept aliases.

Zenthial:
image

Passes all group tests (obviously changed out the cookie, groupid, userids and rolesetids supplied)

suufi:
I'm going to table this for now. As @alanbixby mentioned, this would introduce breaking changes to anyone who uses object parameters.

Neztore:
This would very much be a breaking change, and I can see a few key issues (These have, for the most part, already been identified above):
- Not all the 'target' values are user ids
- "Target" is a unified identifier, even if it is not the clearest on it's own - it's consistent between methods.
- There is no real advantage to switching to user id
- It would be a big breaking change to switch
For these reasons, I am going to close this pull request - if you disagree feel free to join the discord to discuss!

PR #345, opened 2021-05-25
Implement getGroupGames()

alanbixby:
**This is a duplicated fork of #312, but has a readable commit/file history, and has some logic errors fixed.**

~~**NOTE:** A test case was added, but currently it has to point to one of my test groups instead of `noblox.js` as some reason all new games I try to create become personal games instead of group games.. possibly a symptom of their experience/game rushed rebranding.~~ Test now uses the `noblox.js` test group.

Screenshot of the function working:

image

image

### DIFF
--- a/test/group.test.js
+++ b/test/group.test.js
@@ -1,4 +1,4 @@
-const { changeRank, getAuditLog, getGroup, getGroupFunds, getGroupTransactions, getJoinRequests, getLogo, getPlayers, getRankInGroup, getRankNameInGroup, getRole, getRolePermissions, getRoles, getShout, getWall, setRank, shout, setCookie } = require('../lib')
+const { changeRank, getAuditLog, getGroup, getGroupFunds, getGroupGames, getGroupTransactions, getJoinRequests, getLogo, getPlayers, getRankInGroup, getRankNameInGroup, getRole, getRolePermissions, getRoles, getShout, getWall, setRank, shout, setCookie } = require('../lib')
 
 beforeAll(() => {
   return new Promise(resolve => {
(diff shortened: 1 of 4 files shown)


test/group.test.js
```diff
@@ -111,6 +111,27 @@ describe('Group Methods', () => {
     })
   })
 
+  it('getGroupGames() returns an array of group games', () => {
+    return getGroupGames({ groupId: 9997719, limit: 1 }).then((res) => { // FIXME: Convert to noblox.js groupId, currently Roblox broke the creation of group games during the experience rebranding
```
suufi, line 115: Left a fix me.

typings/index.d.ts
```diff
@@ -39,6 +39,17 @@ declare module "noblox.js" {
         IsPrimary: boolean,
     }
 
+    interface GroupGameInfo {
+        id: number;
+        name: string;
+        description: string | null;
```
suufi, line 45: `description?` maybe?

typings/index.d.ts
```diff
@@ -39,6 +39,17 @@ declare module "noblox.js" {
...
+        creator: {id: number; type: string;};
```
suufi, line 46: tad messy, could do `{ id: number, type: string};`
same for L47

typings/jsDocs.ts
```diff
@@ -774,6 +774,20 @@ type Group = {
     isLocked: boolean;
 }
 
+/**
+ * @typedef
+ */
+type GroupGameInfo = {
```
suufi, line 780: see index.d.ts changes

suufi (changes requested):
Looks good for the most part, small things I've noticed.

typings/index.d.ts
alanbixby: Fixed and I corrected the types for some other interfaces that used null unions instead of optional parameters.

suufi (approved):

Vilksian:
I'll close my original pull request.

PR #355, opened 2021-06-05
Define onAuditLog delay (10s) and defaultDelay

alanbixby:
Currently `onAuditLog()` references a delay that does not exist, therefore the delay defaults to `undefined` -> `0`.

This causes many individuals to have `429: TooManyRequest` errors; theoretically this should alleviate things for a fair number of people.

I've also added `defaultDelay` to `settings.json` as a fallback in-case any future event listeners accidentally make this mistake.


Neztore (approved):
Looks good!
The combined use of a ternary and OR are perhaps and difficult to read, but I think this is mergable regardless.

PR #361, opened 2021-06-08
Add setOptions

### LINKED ISSUE #357: setOptions
**Is your feature request related to a problem? Please describe.**
I hate editing the settings.json file everytime I need to change a ratelimit.

**Describe the solution you'd like**
Add a setOptions function to change it for me.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

refer to closed pull request https://github.com/noblox/noblox.js/pull/356 for a code example

Neztore:
Fixes #357 
Adds setOptions, something we've needed for a while. This is a backwards compatible new feature.

This allows settings options to be altered at run-time, rather than requiring a file within the library to be edited.
This is better practice and makes life easier for those using automatic deployment (and plays more nicely with dependency management in general)
This works by validating provided input objects against the existing settings.json file. Entries which exist in the original file are accepted, those that do not are rejected with an error.

The input object must conform to the same structure. In nested objects, omitted values are left unmodified. 

I have not included full typings because I felt due to the extreme size of the settings.json file these would be verbose and unnecessary - it is better to refer directly to the original file.


alanbixby:
image

alanbixby (approved):

suufi (approved):

PR #375, opened 2021-06-29
Add getRobuxBalance()

BrandonRemming:
This would return the Robux balance of the account as a value.


colenh:
you can only get the authorized users robux

alanbixby:
@BrandonRemming This functionality already exists via `getCurrentUser('RobuxBalance')`

PR #376, opened 2021-06-30
Simplify jar overloading by permitting string inputs

alanbixby:
To overload `jar`, you need to pass an object `{ session: 'COOKIE' }`, this is not intuitive and results in over complicated looking code.

```js
await noblox.getCurrentUser({ jar: { session: 'COOKIE' } })
// is now equivalent to
await noblox.getCurrentUser({ jar: 'COOKIE' })
```

This pull request removes the need for the session object, enabling a cookie string to be passed instead of the wrapper object- **backwards compatibility is maintained.**

---

`string` and `object (str)` highlight the new functionality

image
image

### DIFF
--- a/lib/util/getCurrentUser.js
+++ b/lib/util/getCurrentUser.js
@@ -35,7 +35,13 @@ exports.func = function (args) {
         throw new Error('You are not logged in.')
       } else {
         const json = JSON.parse(res.body)
-        return (option ? json[option] : json)
+        if (option) {
+          const searchKey = Object.keys(json).filter((key) => {
+            return option.toLowerCase() === key.toLowerCase()
+          })[0]
+          return json[searchKey]
+        }
+        return json
       }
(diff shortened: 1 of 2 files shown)


alanbixby:
When making the test case screenshots I also noticed `getCurrentUser()` is case-sensitive, so I slapped a quick filter in to patch that too. 

```js
await noblox.getCurrentUser()
await noblox.getCurrentUser('UserName')
await noblox.getCurrentUser('uSeRNaME') // <- used to return undefined
```

**OLD:**
image
**NEW:**
image

Neztore (approved):
I initially had reservations as this would not play well with setting `session_only` to false.
As it turns out, setting session_only to false is broken in the current version of the library anyway - is it worth supporting going forward? I think this change is fine and simplifies authentication in a way which is closer to what we want

PR #377, opened 2021-07-04
Fix getAwardedTimestamps typings, jest case, allow number badgeId- breaking change

alanbixby:
This pull request was intended to make `badgeId` accept a `number` **or** `Array<number>` instead of just `Array<number>`, however I stumbled into several blatant errors that were also fixed.

```js
// New Functionality
await noblox.getAwardedTimestamps(2416399685, [2124549302])
// is now equivalent to
await noblox.getAwardedTimestamps(2416399685, 2124549302)
```

In it's previous state `getAwardedTimestamps` needlessly returns a data object with no cursors, this is not reflected in [`UserBadgeStats`](https://noblox.js.org/global.html#UserBadgeStats), likewise, `awardedDate` is currently a `string`, not a `Date`:
```js
// Old Return
{
    data: [ { badgeId: 19450168, awardedDate: '2010-01-09T20:54:57.38Z' } ]
}
// New Return
[ { badgeId: 19450168, awardedDate: 2010-01-09T20:54:57.38Z } ]
```

I have fixed these errors, updated the return type, and made the jest case more precise to reflect these changes.

image

---

This pull request introduces two breaking changes:
- The return is no longer encapsulated by a `data` object - QoL improvement which agrees with the type interface, but introduces a breaking change
- `awardedDate` is now a `Date` instead of a `string` - this is what the original typings intended
- The parameter `badgeId` is renamed to `badgeIds` - breaks object parameter (included since return already would be breaking)

### DIFF
--- a/lib/badges/getAwardedTimestamps.js
+++ b/lib/badges/getAwardedTimestamps.js
@@ -2,7 +2,7 @@
 const http = require('../util/http').func
 
 // Args
-exports.required = ['userId', 'badgeId']
+exports.required = ['userId', 'badgeIds']
 
 // Docs
 /**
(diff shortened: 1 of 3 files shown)


Neztore (commented):
These changes look good.
I am, however, not sure if it is worth triggering a breaking change for the sake of making a parameter name plural.

Is there a way we could do this while supporting both badgeId and badgeIds? is the increase in library usability worth it? This would be a Semver major change.

Neztore:
> These changes look good. I am, however, not sure if it is worth triggering a breaking change for the sake of making a parameter name plural.
> 
> Is there a way we could do this while supporting both badgeId and badgeIds? is the increase in library usability worth it? This would be a Semver major change.

is it worth revisiting this?

PR #383, opened 2021-07-09
Fix getPlayerInfo() - remove status

alanbixby:
This pull request removes `status` from the return type of `getPlayerInfo()` which was already nearing deprecation status. This may not be favorable as it introduces a breaking change for those reliant on status in `getPlayerInfo()`.

See #384 for an alternative.

---

**Context:** As of July 9th, 2021, the endpoint `https://users.roblox.com/v1/users/<USERID>/status` developed a rate limit around 1 request per minute- which can be considered unusable for most applications.

### DIFF
--- a/lib/user/getPlayerInfo.js
+++ b/lib/user/getPlayerInfo.js
@@ -21,7 +21,6 @@ function getPlayerInfo (userId) {
   return new Promise((resolve, reject) => {
     const requests = [
       constructRequest(`//users.roblox.com/v1/users/${userId}`),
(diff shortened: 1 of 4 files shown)


zomien:
Doesn't make sense to remove it because of a ratelimit change, the endpoint still exists... People may still want to use it. A better solution imo would be to have a parameter in getPlayerInfo that accepts what data someone would want to retrieve from the Roblox user, but with statuses by default left out. So they would have to explicitly want statuses to be included for the request to be made.

alanbixby:
Closing since the upstream rate-limiting issue appears to have resolved itself.

PR #384, opened 2021-07-09
Fix getPlayerInfo() - use www endpoint

alanbixby:
This pull request bypasses the rate limit issues by transitioning from the `users.` endpoint to a `www.`; this not favorable as Roblox does not warn of changes on these endpoints, however this pull request does not introduce any breaking changes.

Since the `www.` endpoint is being implemented, this simplifies the method quite a bit, removing four redundant web requests.

See #383  for an alternative.
 
---

**Context:** As of July 9th, 2021, the endpoint `https://users.roblox.com/v1/users/<USERID>/status` developed a rate limit around 1 request per minute- which can be considered unusable for most applications.

### DIFF
--- a/lib/user/getPlayerInfo.js
+++ b/lib/user/getPlayerInfo.js
@@ -1,6 +1,5 @@
 // Includes
 const http = require('../util/http.js').func
-const getPageResults = require('../util/getPageResults.js').func
 
 // Args
 exports.required = ['userId']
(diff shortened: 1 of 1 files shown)


Chosey98 (approved):
Works perfectly and is a good bypass around the ratelimit introduced today on 1 request per minute for the /status endpoint

alanbixby:
Closing since the upstream rate-limiting issue appears to have resolved itself.

PR #391, opened 2021-07-20
Deprecate the follow method & update test functions

suufi:
- The `follow` method is currently under CAPTCHA and this deprecates the function and marks it as subject to removal in a future version.
- Added `getUAIDs` and `getPremium` to the user.test.js file
- Updated `setRank` so that it no longer errors when testing (attempting to set a user's rank to the same thing results in an error now). 
- Removed `login` from typings because it no longer exists


Neztore (changes requested):
The follow.js file has been updated, but the typings file has not. afaik this means the docs will show deprecated but hints/linters will not catch it?

Neztore (approved):

PR #403, opened 2021-08-06
add onGroupTransaction

### LINKED ISSUE #307: onGroupTransaction
# Steps:
1. Add shortPoll method like the other events
2. Make getGroupTransaction return a bluebird promise

Also should we consider stopping using bluebird given how widely used native promises are?
Either way I think we need a consistent approach where it's either **always** a bluebird promise, or **always** a native promise - instead of a mixture.

Neztore:
Implement onGroupTransaction, resolves #307 

Interval is set to 60 sec due to issues with very low rate limits.


lib/group/onGroupTransaction.js
```diff
@@ -0,0 +1,62 @@
...
+exports.required = ['groupId']
+exports.optional = ['jar']
+
+// Docs
+/**
+ * 🔐 An event for when a group transaction is made, for example a purchase. This event has a rate of one request per 60
+ * sec, which is more than the typical 10 seconds. This is due to the unusually low rate limit on the transactions
+ * endpoint.
+ * @category Group
+ * @alias onShout
```
suufi, line 15: Incorrect alias

suufi (changes requested):

Neztore:
Merged in other PR

PR #404, opened 2021-08-06
Implement searchGroups

### LINKED ISSUE #379: searchGroup Function
Is your feature request related to a problem? Please describe.
A way to have a group search feature with noblox using the [group search](https://groups.roblox.com/docs#!/GroupSearch/get_v1_groups_search) endpoint

Describe the solution you'd like
A function to search for a group, like this
```js
const noblox = require("noblox.js");

noblox.searchGroup('Admin', 10) //10 is the default limit, the limits are 10, 25, 50 and 100
```

Neztore:
Resolves #379

### DIFF
--- a/lib/group/searchGroups.js
+++ b/lib/group/searchGroups.js
@@ -12,7 +12,8 @@ exports.optional = ['prioritizeExactMatch', 'limit']
  * @alias searchGroups
  * @param {string} keyword - The keyword or search term to search by.
  * @param {boolean} [false] prioritizeExactMatch - Whether or not to prioritize the exact match for the keyword
- * @param {Limit=} [limit=∞] - The maximum number of groups to return
+ * @param {number} [100] - The maximum number of groups to return. Supply 0 to apply no limit, returning all results.
+ * Note: This may significantly increase the time to execute and return potentially thousands of results.
  * @returns {Promise<GroupSearchItem[]>}
  * @example const noblox = require("noblox.js")
  * const groupInfo = await noblox.searchGroups("noblox.js");
@@ -23,7 +24,8 @@ exports.func = async function (args) {
   const results = await getPageResults({
     url: '//groups.roblox.com/v1/groups/search',
     query: { keyword: args.keyword, prioritizeExactMatch: args.prioritizeExactMatch || false },
-    limit: args.limit
+    // If limit is 0, pass undefined for infinite results. Otherwise, default to 100.
+    limit: args.limit === 0 ? undefined : (args.limit || 100)
   })
   // Parse updated/created to Date objects.
   return results.map((g) => ({ created: new Date(g.created), updated: new Date(g.updated), ...g }))
--- a/lib/group/searchGroups.js
+++ b/lib/group/searchGroups.js
@@ -11,7 +11,7 @@ exports.optional = ['prioritizeExactMatch', 'limit']
  * @category Group
  * @alias searchGroups
  * @param {string} keyword - The keyword or search term to search by.
- * @param {boolean} [true] prioritizeExactMatch - Whether or not to prioritize the exact match for the keyword
+ * @param {boolean} [false] prioritizeExactMatch - Whether or not to prioritize the exact match for the keyword
  * @param {Limit=} [limit=∞] - The maximum number of groups to return
  * @returns {Promise<GroupSearchItem[]>}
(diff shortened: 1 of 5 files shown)


lib/group/searchGroups.js
```diff
@@ -0,0 +1,31 @@
...
+ * @example const noblox = require("noblox.js")
+ * const groupInfo = await noblox.searchGroups("noblox.js");
+ **/
+
+// Define
+exports.func = async function (args) {
+  const results = await getPageResults({
+    url: '//groups.roblox.com/v1/groups/search',
+    query: { keyword: args.keyword, prioritizeExactMatch: args.prioritizeExactMatch || false },
+    sortOrder: args.sortOrder || 'Asc',
```
suufi, line 26: No sort order actually

suufi (changes requested):

lib/group/searchGroups.js
```diff
@@ -0,0 +1,31 @@
...
+ * @returns {Promise<GroupSearchItem[]>}
+ * @example const noblox = require("noblox.js")
+ * const groupInfo = await noblox.searchGroups("noblox.js");
+ **/
+
+// Define
+exports.func = async function (args) {
+  const results = await getPageResults({
+    url: '//groups.roblox.com/v1/groups/search',
+    query: { keyword: args.keyword, prioritizeExactMatch: args.prioritizeExactMatch || false },
```
alanbixby, line 25: This is inconsistent with the JSDocs; `prioritizeExactMatch` should default to true.

lib/group/searchGroups.js
```diff
@@ -0,0 +1,31 @@
...
+exports.optional = ['prioritizeExactMatch', 'limit']
+
+// Docs
+/**
+ * ✅ Searches for groups by a given search term.
+ * @category Group
+ * @alias searchGroups
+ * @param {string} keyword - The keyword or search term to search by.
+ * @param {boolean} [true] prioritizeExactMatch - Whether or not to prioritize the exact match for the keyword
+ * @param {Limit=} [limit=∞] - The maximum number of groups to return
```
alanbixby, line 15: What behavior does this manifest? Do most search terms go >10,000 results? Perhaps we should enforce an arbitrary default limit i.e. `100`, and let `false` correspond to infinity- otherwise we may see issues of people with infinite loops (in the practical sense)

test/group.test.js
```diff
@@ -105,6 +105,21 @@ describe('Group Methods', () => {
     })
   })
 
+  it('searchGroups() returns groups that match the query', () => {
+    return searchGroups('noblox.js').then((res) => {
```
alanbixby, line 109: Follow-up regarding my limit concern; won't this test run far longer than it needs to?

typings/index.d.ts
```diff
@@ -798,6 +798,17 @@ declare module "noblox.js" {
...
     }
 
+    interface GroupSearchItem
+    {
+        id: number;
+        name: string;
+        description: string;
+        memberCount: number;
+        publicEntryAllowed: boolean;
+        created: Date
```
alanbixby, line 808: Missing semi-colons.

typings/jsDocs.ts
```diff
@@ -952,6 +952,19 @@ type Group = {
...
+/**
+ * @typedef
+ */
+type GroupSearchItem = {
+    id: number;
+    name: string;
+    description: string;
+    memberCount: number;
+    publicEntryAllowed: boolean;
+    created: Date
```
alanbixby, line 964: Missing semi-colons.

Neztore: There are a lot of lines in both typings files without semicolons - Is this something required by typescript? Generally in the library we conform to Standardjs, which prohibits semicolons.

lib/group/searchGroups.js
Neztore: I agree; I'll admit I had not looked enough into the limiting behaviour as I was allowing prioritizeExactMatch to default to true.

I suggest a default limit of 100, but with the ability to supply `0` as no limit, so it's still of the same type.

lib/group/searchGroups.js
Neztore: Altered JsDoc; Defaulting to false makes more sense and matches the endpoint's behaviour

suufi (approved):

PR #419, opened 2021-08-23
Fix overloaded jar in CSRF and wss based methods

alanbixby:
Patches an oversight from #376 which presumed all `jar` passed through `http()`.

`getGeneralToken()` relies on `getHash()` which expects `{ session: 'COOKIE' }`, thus breaking overloaded string jars in all CSRF based methods. Likewise, `onNotification()` manually populates the header, and also needed the conversion.

This fixes those cases, and _does not_ introduce breaking changes.


suufi (approved):

Neztore:
Test on this PR is bugged. I have run locally & had no issues.
Force merging to get around this bug

PR #432, opened 2021-09-27
Support for getDisplaynameFromId and update to getUsernameFromId

conman0552:
Basically the changes outlined are exactly what's in the title. Modified getUsernameFromId to pull from a different endpoint, and then used the same endpoint on another function to create getDisplaynameFromId.

### DIFF
--- a/lib/user/getDisplaynameFromId.js
+++ b/lib/user/getDisplaynameFromId.js
@@ -1,0 +1,44 @@
+// Includes
+const http = require('../util/http.js').func
+const cache = require('../cache')
+
+// Args
+exports.required = ['id']
+
+// Docs
+/**
+ * ✅ Get a user's displayname from their user id.
+ * @category User
+ * @alias getDisplaynameFromId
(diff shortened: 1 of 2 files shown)


alanbixby:
`getUsernameFromId()` is largely considered a legacy method (especially considering its usage of `api.`).

`displayName` can be retrieved via [`getPlayerInfo()`](https://noblox.js.org/global.html#getPlayerInfo), using the same endpoint. Against my opinion, I believe the library is currently leaning away from duplicating it's attributes into multiple redundant methods, i.e. `getDisplayNameFromId`, `getAgeFromId`, and so on.

conman0552:
So would this pull request be more suitable if all the `api.` mentions are converted over to the newer apis (where possible), and the `getDisplayNameFromId` is just dropped?

Neztore:
I agree with Alan here in that it's best to have a single method which returns this and other information in a single object, where that information is still retrieved with one request.

PR #451, opened 2021-10-18
Implement onGroupTransaction

### LINKED ISSUE #307: onGroupTransaction

suufi:
Authored by @Neztore
Implement onGroupTransaction, resolves #307
Interval is set to 60 sec due to issues with very low rate limits.


Neztore (approved):
lovely stuff

FxllenCode:
This is amazing, cheers!

PR #466, opened 2021-11-14
Bumped the getGroupTransactions url to v2

firsttobebear:
Current getGroupTransactions is not working with the cursor parameter but bumping it up to v2 solves the issue. Roblox is now using this by default.

Everything remains the same except for the URL.

### DIFF
--- a/lib/group/getGroupTransactions.js
+++ b/lib/group/getGroupTransactions.js
@@ -1,8 +1,7 @@
-const http = require('../util/http.js').func
-const Promise = require('bluebird')
+const getPageResults = require('../util/getPageResults.js').func
 
 exports.required = ['group']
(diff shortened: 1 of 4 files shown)


Neztore (dismissed):
Also fixes some issues with the current method returning an id of 0 for all transactions.

Neztore:
Just noticed that `id` isn't in the docs. Could you please update docs to match the return type.

I note the typings are already up to date.

Neztore (changes requested):
Please add `id` to docs

Neztore (approved):

Neztore:
Just noticed this PR breaks onGroupTransaction - the version using getPageResults does not play nicely with it

PR #480, opened 2021-12-04
Added typings for updateUniverseAccess function

LightYagami200:


Neztore:
This method does not take in a token; It is fed on internally.
Line 50:
https://github.com/noblox/noblox.js/blob/ac4ca01d7274f4da00023b927e164111a30f2f63/lib/game/updateUniverseAccess.js#L50

Defines the 'public' method here, and it takes universeId, isPublic and jar.

Jar is also an optional parameter, as if it is not specified we fall back onto the internally held jar.
So please:
- Make jar optional
- Remove the token parameter from typings

Thanks for the PR!

Neztore (changes requested):
This method does not take in a token; It is fed on internally.
Line 50:
https://github.com/noblox/noblox.js/blob/ac4ca01d7274f4da00023b927e164111a30f2f63/lib/game/updateUniverseAccess.js#L50

Defines the 'public' method here, and it takes universeId, isPublic and jar.

Jar is also an optional parameter, as if it is not specified we fall back onto the internally held jar.
So please:
- Make jar optional
- Remove the token parameter from typings

Thanks for the PR!

Neztore:
Fixed in #505 , should have been closed ages ago

PR #490, opened 2021-12-29
add presenceType to FriendEntry

policetonyR:
API returns presenceType, why not add it to typings?


Neztore (changes requested):
Looks good, please add `presenceType` to the jsDoc typings too.

policetonyR:
will do when i wake up

Neztore (approved):

PR #492, opened 2021-12-30
fixed shortpoll issue

amazman:
just removed the error event and problem solved. it will error with the socket hang up but that seems to be common but shortpolling will work once again


Neztore:
Thanks for your PR. This does not solve the underlying problem. Hiding errors from the user/not reporting them is not a fix

PR #496, opened 2022-01-04
Update dependencies

Neztore:
Tried to resolve some security vulnerabilities, but didn't get too far. Nested dependencies are a problem; we might need to fork better-docs.

Noblox is not vulnerable, but having 3 "high" security vulnerabilities looks bad, even if they don't apply to this library.


Neztore:
Assigning this minor to also trigger a release

Issue #461, opened 2021-11-12
allow group description changes


Aerosphia:
a noblox function which allows the change of group descriptions would be nice. something like `changeDesc(string: description)` is it all that's needed

thefakewater:
Seems like a good feature.
Congrats on the first issue, btw

PR #503, opened 2022-01-12
Add setGroupDescription and setGroupName

Neztore:
Add setGroupDescription and setGroupName. These are clones of shout - I intend to reduce the amount of duplicate code here in a future version when we move to a different http library and centralise a lot of code.

Implements #461

### DIFF
--- a/typings/jsDocs.ts
+++ b/typings/jsDocs.ts
@@ -1067,7 +1067,7 @@ type GroupDescriptionResult = {
  * @typedef
  */
 type GroupNameResult = {
-    newDescription: string
+    newName: string
 }
 
 /**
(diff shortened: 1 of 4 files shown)


typings/jsDocs.ts
```diff
@@ -1056,6 +1056,20 @@ type GroupShout = {
...
+ */
+type GroupDescriptionResult = {
+    newDescription: string
+}
+
+/**
+ * @typedef
+ */
+type GroupNameResult = {
+    newDescription: string
```
suufi, line 1070: newDescription?

suufi (changes requested):

typings/jsDocs.ts
Neztore: Oops. Forgot to update JSDOC as well as the typings.

suufi (approved):

Issue #431, opened 2021-09-27
Socket Hang Up on onShout [BUG]


conman0552:
![image](https://user-images.githubusercontent.com/24281023/134835542-9eeb66cb-e69e-45d1-a767-087c640498d7.png)
Attached is the image of the issue that keeps reoccurring. After this issue is thrown. the bot no longer polls the server.

It looks like getShout is handling the error fine, but onShout.js at line 48 is missing a .catch statement, which it should have given that it's possible for getShout to reject the promise.

To reproduce, all I've needed to do is run noblox with the onShout event, and let it sit. After a bit the error pops up.

conman0552:
After going through the code, it looks like [this issue](https://github.com/noblox/noblox.js/issues/429) is also caused potentially by the same thing, no error catching in the shortpoll. If I had to guess anything that uses shortpoll is probably effected.

Neztore:
Code?

Issue #429, opened 2021-09-22
[BUG] Socket hang up error when using the onWallPost event


Brandong04:
**Describe the bug**
This issue is regarding the "socket hang up" error that's floated around on several occasions when people use the onWallPost event. This issue has been present for a while, and I'm hoping this will help bring it to a close at last. 

First and foremost, from my experimenting with this, it appears to be completely random as to when it happens, but typically I can have it happen a few times within about 5 minutes.

A few other things to note are that after some back and forth with some other people, it can also be dependent on internet speeds (which I by no means have slow internet, so I'm not sure on that one).

Handling this through the on error event only catches the SubError, rather than both. So at the moment from my knowledge, it's also not feasible to handle this manually without some extra intervention. 

**To Reproduce**
1. Use noblox.js' onWallPost event
2. Wait a random period of time and a "socket hang up" error will show.

**Expected behavior**
The onWallPost event should function normally, and not randomly throw socket hang up errors without explanation. Handling these errors through the event should also catch this error, and not just the SubError (more about this in the additional context section).

**Screenshots**
Basic code example:
![image](https://user-images.githubusercontent.com/37907304/134360410-1b32a322-4ae5-422c-8e9b-7653419d89fb.png)

Error Example:
![image](https://user-images.githubusercontent.com/37907304/134360773-365930c5-3ec6-4934-833f-5d3bc69d83ce.png)

**Additional context**
Handling errors through the wallPost.on('error', ()  => {}) event will only handle the SubError and not the socket hang up, the SubError will appear after the initial socket hang up error.

Example of this: 
![image](https://user-images.githubusercontent.com/37907304/134361877-428a8160-c862-47d3-a0fe-67511d91cc4a.png)

WaviestBalloon:
I'm running into this issue, bumping

WaviestBalloon:
Error output:
```javascript
Unhandled rejection RequestError: Error: socket hang up
    at new RequestError (G:\Desktop\bot\node_modules\request-promise-core\lib\errors.js:14:15)
    at Request.plumbing.callback (G:\Desktop\bot\node_modules\request-promise-core\lib\plumbing.js:87:29)
    at Request.RP$callback [as _callback] (G:\Desktop\bot\node_modules\request-promise-core\lib\plumbing.js:46:31)
    at self.callback (G:\Desktop\bot\node_modules\request\request.js:185:22)
    at Request.emit (events.js:400:28)
[...]

PR #504, opened 2022-01-12
Add onJoinRequest, re-document onJoinRequestHandle, Fix uncatchable errors and drop bluebird

### LINKED ISSUE #488: [BUG] onJoinRequestHandle does not behave as is documented
Currently using this excerpt of code as an event handler for `onJoinRequestHandle`:
```js
// Special Event Handler
(async () => {
    await noblox.setCookie(process.env.cookie).catch(console.error);
    const onJoinRequestHandle = require(`../src/events/onJoinRequestHandle`);

    noblox.onJoinRequestHandle(config.group).on("data", (...args) => onJoinRequestHandle(mongoClient, ...args));
})().catch(console.error);
```

Receiving some issues though. This doesn't seem to fire unless I restart my hosting process. When restarting my hosting process, it fires for all members pending a join request (immediately), but never continues to do so until I restart again.

This may or may not be an internal issue, but I received a bit of confirmation [here](https://discord.com/channels/797610307928326194/797610307928326199/922994710325374976) (you can see the full Discord conversation there too).

If it indeed is an internal issue, can I get an ETA on a fix please? Would appreciate it.

Neztore:
Normally, I wouldn't include so many changes in a single PR. But this one spiralled a little, and each little thing ended up being connected.

Firstly, this PR adds `onJoinRequest`. This behaves much like onAuditLog and onShout - it emits whenever more 'recent' items appear. It acts in the way that `onJoinRequestHandle` was documented to act. In my opinion it is also a better design pattern to use the event to emit and a separate library function to act upon those emissions.

Secondly, this PR adds documentation for `onJoinRequestHandle` from the [original wiki](https://github.com/sentanos/roblox-js/wiki/Group-Functions#onjoinrequesthandle) which describes it's true functionality.

Thirdly, This PR drops bluebird and adds a Polyfill method in `internal` `Timeout`. This is because async functions are not bluebird functions, which means `Promise.timeout` does not exist. This was causing issues with the shortPoll function.
Bluebird is no longer needed beyond the utility functions it provides - and as far as I am aware, timeout is the only one we use.
Native promises are now faster. I'm quite pleased with this part.

Finally, and this is a big one, this PR fixes some uncatchable errors. These were especially prevalent with events. This is primarily due to a design pattern within the library of wrapping promises (usually http promises) in an outer, manually created promise. This looks like:
```js
const outer = new Promise(function (resolve, reject) {
  return innerWillReject()
    .then(function (v) {
      // ...
      resolve(v)
    })

})
  .then(function (v) {
    console.log(`Result: ${v}`)
  })
  .catch(function (e) {
    console.log(`Error: ${e}`)
  })

function innerWillReject () {
  return new Promise(function (resolve, reject) {
    reject("Error!")
  })
}
```

The issue with this is that when the inner function rejects, it does not propagate up the promise chain and instead becomes an unhandled rejection - that cannot be handled. This error was hard to spot because it was handled for *some* methods, and it was only noticeable when an underlying http error occurred, such as the infamous socket hang-up from events. I have tested and these events are now passed correctly to `error`, and emitted from the event. Quite regularly in some cases, but that's unavoidable. 

Resolves #488 #431 #429

### DIFF
--- a/lib/group/onJoinRequest.js
+++ b/lib/group/onJoinRequest.js
@@ -1,0 +1,63 @@
+// Includes
+const settings = require('../../settings.json')
+const shortPoll = require('../util/shortPoll.js').func
+const getJoinRequests = require('./getJoinRequests.js').func
+const promiseTimeout = require('../internal/timeout')
+
+// Args
+exports.required = ['group']
+exports.optional = ['jar']
+
+// Docs
+/**
+ * 🔐 An event for when someone makes a request to join the group.
+ * @category Group
+ * @alias onJoinRequest
+ * @param {number} group - The id of the group.
+ * @returns An EventEmitter that emits when someone tries to join.
+ * @example const noblox = require("noblox.js")
+ * // Login using your cookie
+ * const e = noblox.onJoinRequest()
+ * e.on("data", function(data) {
+ *  console.log("New request!", data)
+ * })
+ * e.on("error", function (err) {
+ *   ...
+ * })
+ **/
+
+async function getRequests (jar, group, cursor) {
+  const requests = []
+  const res = await promiseTimeout(getJoinRequests({ jar, group, cursor, limit: 100 }), settings.event.timeout, 'getReqeusts onJoinRequests internal')
+  requests.push.apply(requests, res.data)
+  if (res.nextPageCursor) {
+    requests.push.apply(requests, await getRequests(jar, group, res.nextPageCursor))
+  }
+  return requests
+}
+
+// Define
+exports.func = function (args) {
+  return shortPoll({
+    getLatest: function (latest) {
+      return getRequests(args.jar, args.group)
+        .then(function (requests) {
+          const given = []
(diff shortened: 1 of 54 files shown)


lib/group/onJoinRequest.js
```diff
@@ -0,0 +1,63 @@
...
+ *  console.log("New request!", data)
+ * })
+ * e.on("error", function (err) {
+ *   ...
+ * })
+ **/
+
+async function getRequests (jar, group, cursor) {
+  const requests = []
+  const res = await promiseTimeout(getJoinRequests({ jar, group, cursor, limit: 100 }), settings.event.timeout, 'getReqeusts onJoinRequests internal')
```
suufi, line 31: getRequests*

suufi (changes requested):

suufi:
Regenerated yarn.lock because of how problematic it is.

suufi (approved):

PR #505, opened 2022-01-13
Add updateUniverse & getUniverseInfo

popeeyy:
Add updateUniverse and the respective typings. Also properly fixed updateUniverseAccess typings (#480).
(BREAKING) Replace getPlaceInfo (uses roblox.com endpoint) with getUniverseInfo.


lib/game/updateUniverse.js
```diff
@@ -0,0 +1,53 @@
...
+ * @returns {Promise<UpdateUniverseResponse>}
+ * @example const noblox = require("noblox.js")
+ * // Login using your cookie
+ * noblox.updateUniverse(2421261122, { name: "The best game ever!" })
+**/
+
+// Define
+function updateUniverse (universeId, settings, jar, token) {
+  return new Promise((resolve, reject) => {
+    return http({
```
Neztore, line 25: This falls into the anti-pattern frequently present within the library - The http promise is not caught, as it is within a Promise which has been manually created. Please add .catch to this promise!

Neztore (changes requested):

typings/index.d.ts
```diff
@@ -733,6 +733,98 @@ declare module "noblox.js" {
...
+        universeAvatarMaxScales?: AvatarScale
+
+        studioAccessToApisAllowed?: boolean;
+        permissions?: UniversePermissions;
+
+        optInRegions?: Array<Regions>;
+        optOutRegions?: Array<Regions>;
+    }
+
+    interface UpdateUniverseResponse
```
Neztore, line 794: Could this interface be re-written as an [subtype](https://www.typescriptlang.org/docs/handbook/interfaces.html#extending-interfaces) of UniverseSettings. Unless I have missed anything it is identical except it also has 'id'.

Neztore: I note optOutRegions is not present. I'm also not seeing that in the [docs](https://develop.roblox.com/docs#!/UniverseSettings/patch_v1_universes_universeId_configuration)? Is that an undocumented value?

Neztore (changes requested):

typings/index.d.ts
popeeyy: This data is compiled from the data used to create the docs page. The model for this endpoint included it, so I added it in case they decide to use it in the future.

popeeyy: I knew there was something it was forgetting and it was subtypes 😂 I'll have to experiment with better docs to see how it reacts to a subtype in its typings file.

Neztore (approved):

lib/game/getUniverseInfo.js
```diff
@@ -0,0 +1,49 @@
...
+
+exports.required = ['universeIds']
+exports.optional = ['jar']
+
+// Docs
+/**
+ * 🔓 Get the info for a universe.
+ * @category Game
+ * @alias getUniverseInfo
+ * @param {number} universeIds - The ids of the universes.
```
suufi, line 11: Ids implies plural so the parameter should suggest multiple IDs can be used as in the documentation.

suufi (changes requested):

suufi (approved):

PR #534, opened 2022-02-16
Remove Disabled Status Endpoint

Dominick-P:


Chosey98:
I was just about to make a PR for this as it broke the function completely for my bot. Nice PR please merge asap

Chosey98:
Turns out #535 includes the other typing as well so I would suggest just going for that one since duplicate PRs while the other one includes more typings

PR #535, opened 2022-02-17
Remove status from getPlayerInfo

ItsWHOOOP:
Remove status from getPlayerInfo as it is causing issues because Roblox has sunsetted the status endpoint


alanbixby (approved):
image

Functional change, fixes getPlayerInfo.

PR #543, opened 2022-03-09
Remove Authentication

realCartar:
Removed X-CSRF-TOKEN & Cookie Jar from the Request.
I've tested on multiple different occasions that this endpoint doesn't require Authentication.

(I don't know if it used to, but upon my testing, I didn't run into any issues).

### DIFF
--- a/lib/presence/getPresences.js
+++ b/lib/presence/getPresences.js
@@ -1,5 +1,5 @@
 // Includes
-const http = require('../util/http.js').func;
+const http = require('../util/http.js').func
 
 // Args
 exports.required = ['userIds']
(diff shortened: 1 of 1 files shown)


alanbixby:
I believe the authentication headers are required for getting privacy protected presence data. For example if someone had their join-game permissions set to `Friends Only`, then a public request returns `rootPlaceId=0` whereas an authenticated friend would show the placeId they are in.

If that is the case, it should be relabeled from 🔐 (authentication required) to 🔓 (authentication optional) in the docs; and perhaps only the CSRF token is superficial (which breaks the mold from other POST requests).

realCartar:
Superseded by PR #544.

Thanks Alan!

PR #544, opened 2022-03-10
fix: set optional authentication in getPresences()

alanbixby:
Supercedes PR #543.

The CSRF token has been removed but the cookie is necessary to access private game data; documentation has also been updated to reflect this change.

image


alanbixby:
Looking at this again; can I get away with introducing a breaking change so this returns an array instead of a single attribute object of an array- like it should have been in the first place? 

👉👈

PR #546, opened 2022-03-13
make getPresences return an array directly

Neztore:
Features a slightly... dubious... hack for backwards compatibility.

Proof of concept:
```js
const arr = [1,2,3,4];
arr.foo = "bar";
arr.push(5);

for (const i of arr) {
  console.log(i);
}
```

image

This key exists soley for backwards compatibility, and is intentionally not in the typescript typings. Do not use it for new work; it should be removed, if merged, in the next semver major version.

### DIFF
--- a/lib/presence/getPresences.js
+++ b/lib/presence/getPresences.js
@@ -50,11 +50,11 @@ exports.func = function getPresences ({ userIds, jar }) {
           }
           reject(new Error(error))
         } else {
-          const d = responseData.userPresences;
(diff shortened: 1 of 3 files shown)


Neztore:
As above image demonstrates, while this works in a for... of loop it produces some funky behaviour when used with a for in loop.

Perhaps this is not a problem: A for in loop is for iterating over Object properties, after all.
Still not sure if that's sufficient justification for such odd behaviour from a user perspective, though

PR #556, opened 2022-04-05
Adding await to uploadItem to JSDoc function example

athammer:
I know everyone loves small documentation edits 😄. There was some confusion in [discord](https://discord.com/channels/797610307928326194/842966507826970624/960895145971167322) due to this not having an await in the JSDoc example. Adding that to prevent that.


zachariapopcorn:
Does uploadItem even work anymore?

athammer:
good question; let me ask the person who was fiddling with it. I can check when I have more time. i might work but it might only accepts `.png .jpg .tga or .bmp` files.

suufi (approved):

PR #575, opened 2022-05-29
Refactored setCookies function

nikos467:
I refactored it by removing an unnecessary else statement and by using async await instead of .then catch.


Neztore:
Changes in this PR are also in #579

PR #579, opened 2022-05-30
Refactored getCurrentUser.js and setCookie

nikos467:
Made the export function easier  to read

### DIFF
--- a/lib/user/setCookie.js
+++ b/lib/user/setCookie.js
@@ -27,12 +27,10 @@ exports.func = async function (args) {
     options.jar.session = args.cookie
     return false
   }
-  try{
+  try {
     const res = await getCurrentUser({ jar: { session: args.cookie } })
-    options.jar.session = args.cookie
     return res
-  }
-  catch(error){
+  } catch (error) {
     console.error('Failed to validate cookie: Are you sure the cookie is valid?\nEnsure you include the full cookie, including warning text.')
     throw new Error(error)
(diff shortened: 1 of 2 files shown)


lib/user/setCookie.js
```diff
@@ -18,22 +18,20 @@ exports.optional = ['validate']
...
-      .then(function (res) {
-        options.jar.session = args.cookie
-        return res
-      }).catch(function (e) {
-        console.error('Failed to validate cookie: Are you sure the cookie is valid?\nEnsure you include the full cookie, including warning text.')
-        throw new Error(e)
-      })
+  }
+  try {
+    const res = await getCurrentUser({ jar: { session: args.cookie } })
```
Neztore, line 31: Could `return await getCurrentUser(...);` to remove an unnecessary line and variable.

Neztore: Or, a better solution would be to match the original behaviour and actually set the session token following validation!

Please test your PRs. Reading your code, if validate is enabled the internally stored cookie will never be updated.

Neztore (changes requested):
Please edit your PR to set the internally stored token in jar

Chosey98:
Matter of fact I just reviewed and pushed the changes @Neztore requested to @nikos467 branch and waiting his approval to merge

Neztore:
No activity and fixed elsewhere

PR #584, opened 2022-06-17
Replace api.roblox.com Endpoints

popeeyy:
Replaces all api.roblox.com endpoints that will be removed starting July 15, 2022. Not a breaking change yet since Roblox has so far kept some of the responses from the legacy endpoints.
https://devforum.roblox.com/t/action-needed-sunsetting-apirobloxcom/1796897

Does not include getDeveloperProducts since there is not a comparable endpoint yet and the legacy endpoint is still used by Roblox Studio.

### DIFF
--- a/lib/asset/getProductInfo.js
+++ b/lib/asset/getProductInfo.js
@@ -19,7 +19,7 @@ exports.required = ['asset']
 function getProductInfo (asset) {
   return new Promise((resolve, reject) => {
     const httpOpt = {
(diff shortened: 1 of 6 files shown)


suufi:
I'm actually going to not merge this now. I have a similar set of commits, but with some directory renaming to account for the subdomains. I'm waiting for [some additions](https://github.com/noblox/noblox.js/projects/3) so I can open and merge the PR.

PR #595, opened 2022-07-17
Implemented getUniverseContainingPlace

H3xad3cimalDev:
I saw there was no api implementation for
//api.roblox.com/universes/get-universe-containing-place
so I implemented it. Sorry if this is a bad request since this my first pull request.

### DIFF
--- a/lib/game/getUniverseContainingPlace.js
+++ b/lib/game/getUniverseContainingPlace.js
@@ -26,7 +26,7 @@ function getUniverseContainingPlace (placeId) {
 
   return http(httpOpt)
     .then(function (res) {
-      const body = JSON.parse(res.body);
+      const body = JSON.parse(res.body)
       if ('UniverseId' in body) {
         return body.UniverseId
       }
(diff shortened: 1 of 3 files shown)


brandoge91:
api.roblox.com is getting sunsetted soon i think, so this would only work for a little bit.

H3xad3cimalDev:
> api.roblox.com is getting sunsetted soon i think, so this would only work for a little bit.

so if I implement the same thing without using api.roblox.com would it be better?

brandoge91:
yes

H3xad3cimalDev:
> yes

I don't think I will be able to since look at the APIs roblox doesn't seem like they're going to add a alterative to this endpoint.
I will add a "fix" if I do find a alterative but might as well close this pr.

Aviation-7111:
There is another API endpoint that ROBLOX has [suggested using](https://devforum.roblox.com/t/action-needed-sunsetting-apirobloxcom/1796897#universes-6), which can accomplish the same task and more, with a caveat being you need to be authenticated to use it (from my testing at least).

https://games.roblox.com/docs#!/Games/get_v1_games_multiget_place_details, this can take in multiple `placeId`'s and always returns an array of objects with place data, that have their respective `universeId` that the noblox.js user can utilize. Do think this warrants a new PR if implemented.

alanbixby:
Utilizes deprecated endpoints; otherwise good first attempt at a PR.

PR #630, opened 2022-10-11
Fix FriendRequest type

siim0n:
FriendRequest is a number, not an object containing a number


Neztore (approved):

Neztore (changes requested):
Could you also update the corresponding docs entry?
https://github.com/noblox/noblox.js/blob/master/typings/jsDocs.ts#L1490

siim0n:
Hi, thanks for reviewing this, I just updated the docs too

Neztore (approved):

Neztore:
Thanks for your PR!

PR #639, opened 2022-11-01
Fix addDeveloperProduct.js

Astrism:
HTTP Uri was moved by Roblox.

### DIFF
--- a/lib/game/addDeveloperProduct.js
+++ b/lib/game/addDeveloperProduct.js
@@ -1,7 +1,6 @@
 const http = require('../util/http.js').func
 const getGeneralToken = require('../util/getGeneralToken.js').func
 const checkProductName = require('./checkDeveloperProductName.js').func
-const parser = require('cheerio')
 
 exports.required = ['universeId', 'name', 'priceInRobux']
 exports.optional = ['description', 'jar']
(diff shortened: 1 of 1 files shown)


lib/game/addDeveloperProduct.js
```diff
@@ -28,36 +27,26 @@ const nextFunction = (jar, token, universeId, name, priceInRobux, description) =
...
-            }
+          return {
+            universeId,
+            name,
+            priceInRobux,
+            description,
+            productId: typeof res.body === 'object' ? res.body.id : JSON.parse(res.body).id
           }
         } else {
+          console.error(res)
```
Neztore, line 49: Is there a better way we can output this error? Could it be included as part of the throw?
While the method currently provides a pretty unhelpful error message this could be a good opportunity to fix this and throw the actual error.

Neztore (changes requested):

lib/game/addDeveloperProduct.js
FourAstro: Line 50, check last commit to this pull.
Feel free to name changes to the way it throws the wire
error. I suppose adding status codes.

FourAstro: Or did I forget to remove it again.

Astrism: [View](https://github.com/noblox/noblox.js/pull/639/commits/0653f49fe3d09c82bfb416a3bb460979ec1261a7)

Neztore (approved):
Thanks for updating the error behavior - not sure what you were saying about with latest commits: I just check the "Files changed" tab and then clone down the latest copy of the branch!

Thanks for your contribution.

PR #651, opened 2022-12-14
Add getGameSorts and getGameList

alexop1000:
This adds support for getting the different sorts and the games in them.

### DIFF
--- a/lib/game/getGameList.js
+++ b/lib/game/getGameList.js
@@ -1,0 +1,46 @@
+// Includes
+const http = require('../util/http.js').func
+
+// Args
+exports.required = ['token']
+exports.optional = ['jar', 'maxRows']
+
+// Docs
+/**
+ * 🔓 Get the sorts for a game.
+ * @category Game
+ * @alias getGameList
+ * @param {string} token - The sort token to get.
+ * @param {number} maxRows - The maximum number of rows to return.
+ * @returns {Promise<GameList[]>}
+ * @see [getGameSorts()](global.html#getGameSorts) - can be used to get the token
+ * @example const noblox = require("noblox.js")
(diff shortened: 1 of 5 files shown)


lib/game/getGameList.js
```diff
@@ -0,0 +1,40 @@
...
+
+// Define
+function getGameList (token, maxRows, jar) {
+  return http({
+    url: `//games.roblox.com/v1/games/list?sortToken=${token}&startRows=0&maxRows=${maxRows}&hasMoreRows=true`,
+    options: {
+      json: true
+    }
+  })
+    .then((body) => {
```
alanbixby, line 29: We would want this to be using the pagination logic; currently this would only return the first page.

lib/game/getGameSorts.js
```diff
@@ -0,0 +1,38 @@
...
+// Args
+exports.required = ['gameSortsContext']
+exports.optional = []
+
+// Docs
+/**
+ * ✅ Get the sorts for a game.
+ * @category Game
+ * @alias getGameSorts
+ * @param {string|number} gameSortsContext - The sort context to get.
```
alanbixby, line 13: Would prefer defined enums here; I know its tedious given the modern documentation but they are

"GamesDefaultSorts",
"GamesAllSorts",
"HomeSorts",
"ChatSorts"

hilariously, the "updated" docs, list 0 - 7 but only 0 - 3 work without throwing an HTTP 500; thanks Roblox

alanbixby (changes requested):

alexop1000:
That's true, but it's been 4 months and I'm not fixing that now xD. But yeah the Roblox docs is very misleading. They also still use OpenAPI 2.0 instead of 3.0

Neztore:
Author has no plans to bring this to a mergable state, closing. Feel free to pick it up if you need these functions.

PR #685, opened 2023-03-15
Check for existence of shout before parsing dates

Wolftallemo:
The `shout` property is null for groups that have never set a shout.


suufi (approved):

suufi (changes requested):
Typings should be `shout?: GroupShout`

suufi (approved):

Chosey98:
@suufi I believe this is ready to be merged as this is an important commit blocking an important method. If any changes else are requested can you let me know as I need this commit pushed asap so I would like to help.

PR #690, opened 2023-03-25
Add getGameIcon

Cramber25:
this adds support of getting the icon of a game by its universeid as its not shown in getuniverseinfo api

### DIFF
--- a/lib/thumbnails/getGameIcon.js
+++ b/lib/thumbnails/getGameIcon.js
@@ -1,0 +1,54 @@
+// Includes
+const http = require('noblox.js/lib/util/http.js').func
+
+// Args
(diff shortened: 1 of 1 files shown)


lib/thumbnails/getGameIcon.js
```diff
@@ -0,0 +1,54 @@
+// Includes
+const http = require('noblox.js/lib/util/http.js').func
```
suufi, line 2: not proper import

suufi (changes requested):
You also need to add test cases for this function and typings in the appropriate files.

Neztore:
Stale. Please re-open if you get the chance to implement the above.

PR #691, opened 2023-03-29
Fixed getGroup (cannot read properties of null 'created')

kuphg:
Fixed the getGroup function not working if shout is null


Chosey98 (changes requested):
Can you make it a condition thats sets to null if its not there instead?

PR #641, opened 2022-11-04
Add setDescription.

GestaIt:
Closes #633

This PR will introduce the `setDescription` method. It'll allow the user to edit their accounts' description.

PR #702, opened 2023-04-23
Add setDescription

### LINKED ISSUE #633: Add Set Description
Add a function to change the Player Description.

Neztore:
From #641 
Resolves #633 

Since the PR is so old merging was a real pain (as it also made a lot of minor spacing changes to typings files)
So I've copied out the code into a recent branch. Thanks @GestaIt for the contribution.

### DIFF
--- a/test/users.test.js
+++ b/test/users.test.js
@@ -1,4 +1,4 @@
-const { getBlurb, getIdFromUsername, getPlayerInfo, getUsernameFromId, setCookie } = require('../lib')
+const { getBlurb, getIdFromUsername, getPlayerInfo, getUsernameFromId, setCookie, setDescription } = require('../lib')
 
 beforeAll(() => {
   return new Promise(resolve => {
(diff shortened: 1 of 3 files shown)


test/users.test.js
```diff
@@ -59,6 +59,11 @@ describe('Users Methods', () => {
       })
     })
   })
+  it('setDescription() sets a description', async () => {
+    return setDescription("Hello!").then(res => {
+      return expect(res.description).toEqual(expect.any(String));
```
alanbixby, line 64: We likely want a second test-case with this that is using `getBlurb()` to validate the change actually took place on the website (doing two set requests to watch the change).

On that note; we might also want to consider marking `getBlurb()` as deprecated and aliasing it under `getDescription()` for consistency sake

alanbixby (changes requested):
Blocking for more verbose test cases, so we can catch upstream changes more easily.

test/users.test.js
Neztore: getBlurb is already deprecated, with moving to getPlayerInfo recommended instead.
getPlayerInfo returns description with a "blurb" key, and I now considering it further setDescription could cause confusion with setting group description, though it's fairly obvious from typings & comments.
Could go for `setUserDescription` & add a key to getPlayerInfo that includes description in description instead of blurb (but leaving blurb for backwards compatibility)

PR #703, opened 2023-04-23
Update the default maxThreads to 128

FungusGenerator:
Any server that's not windows 95 can handle this, also power of 2 = better obviously


alanbixby:
N/A; `settings.maxThread` is only referenced by `threaded.js` which has no code-coverage pending removal per #708

PR #728, opened 2023-06-30
Switch from Requests to Axios

### LINKED ISSUE #143: "request has been deprecated" warning when installing noblox.js
**Describe the issue**
When installing noblox.js, this following warning pops up in the install logs:
```
npm WARN deprecated request@2.88.2: request has been deprecated, see https://github.com/request/request/issues/3142
```

**To Reproduce**
1. Install noblox.js
2. See warning

**Expected behavior**
A clean install without any warnings.

**Why this Issue is Important**
A future update of npm may break the install sequence for noblox.js.

royce-mathew:
Axios is a much more advanced library compared to request-promise. It is also a non-deprecated library. I've switched almost all functions from requests to axios. This fixes #143 

This is a work in progress.

### DIFF
--- a/lib/accountinformation/getUserSocialLinks.js
+++ b/lib/accountinformation/getUserSocialLinks.js
@@ -25,14 +25,14 @@ function getUserSocialLinks (userId, jar) {
       resolveWithFullResponse: true
     }
   })
-    .then(({ statusCode, body }) => {
-      const { errors } = JSON.parse(body)
-      if (statusCode === 200) {
-        return JSON.parse(body)
-      } else if (statusCode === 400) {
+    .then(({ status, data: resData }) => {
+      const { errors } = resData
+      if (status === 200) {
+        return resData
+      } else if (status === 400) {
         throw new Error(`${errors[0].message} | userId: ${userId}`)
       } else {
-        throw new Error(`An unknown error occurred with getUserSocialLinks() | [${statusCode}] userId: ${userId}`)
+        throw new Error(`An unknown error occurred with getUserSocialLinks() | [${status}] userId: ${userId}`)
       }
     })
 }
--- a/lib/accountsettings/block.js
+++ b/lib/accountsettings/block.js
@@ -34,15 +34,15 @@ function block (jar, token, userId) {
     }
     return http(httpOpt)
       .then(function (res) {
-        if (res.statusCode === 200) {
+        if (res.status === 200) {
           resolve()
         } else {
-          const body = JSON.parse(res.body) || {}
+          const body = res.data || {}
           if (body.errors && body.errors.length > 0) {
             const errors = body.errors.map((e) => {
               return e.message
             })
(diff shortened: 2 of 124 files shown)


zachariapopcorn:
Just use the native fetch() API that got introduced in Node 18

royce-mathew:
I would love to, this is just a basic port of noblox.js over to axios. I figured axios would be better since we wouldn't really have to parse the data to JSON, and it was essentially almost the same format.

Although I do agree it would've saved me a lot more time if I went with fetch instead since it looks like it uses the similar format as request-promise. 

The main reason I ported it over was because I wanted to use the noblox library in vercel which doesn't support request-promise since it's deprecated.

suufi:
You have some stuff that shouldn't have been changed because of find and replace. Also, it's a little more complex than just changing the request library name. If you attempt to run the tests on this, I'm fairly certain a good chunk will fail.

royce-mathew:
Would it be easier to switch it over to the new fetch() api instead? This version fails around 15 test suites whereas the current noblox release fails 10. I'm not sure if I'm missing a environment variable. I have `COOKIE` and `API_KEY`

zachariapopcorn:
> I would love to, this is just a basic port of noblox.js over to axios. I figured axios would be better since we wouldn't really have to parse the data to JSON, and it was essentially almost the same format.
> 
> 
> 
> Although I do agree it would've saved me a lot more time if I went with fetch instead since it looks like it uses the similar format as request-promise. 
> 
> 
> 
> The main reason I ported it over was because I wanted to use the noblox library in vercel which doesn't support request-promise since it's deprecated. 

Axios is known for its random bugs. I've been told to just use fetch()

Neztore:
> Would it be easier to switch it over to the new fetch() api instead? This version fails around 15 test suites whereas the current noblox release fails 10. I'm not sure if I'm missing a environment variable. I have `COOKIE` and `API_KEY`

I, for one, would prefer fetch - means dropping a dependency, and using a native function. Less chance of support being dropped in future.

Axcelers:
I suggest that we move away from requests to something up to date as it has been leaving connections open and not completing certain requests such as ranking.

Neztore:
> I suggest that we move away from requests to something up to date as it has been leaving connections open and not completing certain requests such as ranking.

These issues tend to be more to do with Roblox misbehaving when responding to requests, but I absolutely agree it is a problem.

We are keen to move away, especially now Node.js has a built-in fetch API. Feel free to open a pull request - unfortunately we are pretty short on time these days.

Neztore:
Since there has been no progress since June and we are leaning towards the native fetch API I am going to close this PR. Please-reopen if work resumes.
Otherwise please direct further conversation about upgrading from request to #143 thanks!

Axcelers:
I suggest we should migrate to fetch(). The API connections keeps getting left open, making noblox.js an unreliable source. My bots have been unreliable due to noblox.js using requests.

PR #751, opened 2023-08-28
Fix addDeveloperProduct not returning a product id

iamEvanYT:
When using the new api, it returns a completely different id. I can then use that id to grab the real product id using a GET api.


Neztore (changes requested):
Please make sure tests and `npm run lint` pass. Once CI passes I will review.

Cheers.

iamEvanYT:
@Neztore there we go.

Ratzifutzi:
When will the changes be fixed?

Neztore (approved):

PR #772, opened 2023-11-14
Fixed onAuditLog not working

commonly-ts:
Added sortOrder argument to return the newest audit logs rather than only fetching the oldest ones, resulting in nothing coming through when using onAuditLog.
Sort order default being ascending resulted in the API request only returning the first 100 or so audit logs from the group.

### DIFF
--- a/lib/groups/onAuditLog.js
+++ b/lib/groups/onAuditLog.js
@@ -30,7 +30,7 @@ exports.func = function (args) {
   let empty = false
   return shortPoll({
     getLatest: function (latest) {
-      return getAuditLog({ group: args.group, jar: args.jar })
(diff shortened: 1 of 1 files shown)


Neztore (approved):

lib/groups/onAuditLog.js
```diff
@@ -30,7 +30,7 @@ exports.func = function (args) {
   let empty = false
   return shortPoll({
     getLatest: function (latest) {
-      return getAuditLog({ group: args.group, jar: args.jar })
+      return getAuditLog({ group: args.group, jar: args.jar, sortOrder: "Desc" })
```
Neztore, line 33: Please change this to use Singlequote and check that an `npm run lint` passes.

Neztore (approved):

PR #778, opened 2024-02-13
 Update updateDeveloperProduct function to new 'apis.' endpoint; removes iconImageAssetId parameter (needs new method call)

alanbixby:
The endpoint currently referenced by `updateDeveloperProduct` is deprecated and responds with a HTTP 404; this change updates the method to match what is used by the website.

This introduces a breaking change since `iconImageAssetId` is no longer supported. I recommend we create a new method for updating the icon rather than nesting the request inside this method; especially since the website is now required a full FormData upload instead of creating a decal and referencing it's `assetId`.


Neztore:
Is this similar to #751 ?

Neztore (approved):

PR #782, opened 2024-03-05
Fixed case sensitivity in getIdFromUsername.js

MilesSRC:
Fixed an issue in noblox.js where the API incorrectly handles usernames that are not in lowercase. When getIdFromUsername() receives a username like "ROBLOX," the API responds with "roblox" in the requestedUsername field. Previously, noblox.js did not account for this case-insensitive behavior, leading to issues in users' code. This has now been addressed.

### DIFF
--- a/lib/users/getIdFromUsername.js
+++ b/lib/users/getIdFromUsername.js
@@ -20,8 +20,6 @@ exports.required = ['username']
 function getIdFromUsername (usernames) {
   usernames = Array.isArray(usernames) ? usernames : [usernames] // cast usernames to array if necessary
 
-  usernames.map((username) => username.toLowerCase())
-
   const httpOpt = {
     url: 'https://users.roblox.com/v1/usernames/users',
     options: {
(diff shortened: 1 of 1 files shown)


Jetfeliks746:
It still isn't working for me. I tried both uppercase and lowercase but the getIdFromUsername still return null.

commonly-ts:
I can confirm that this fix works. Tested it on a bot with a user on the Discord who's username didn't work, but after adding this fix to my code, it worked.

commonly-ts (approved):
I can confirm that this fix works. Tested it on a bot with a user on the Discord who's username didn't work, but after adding this fix to my code, it worked.

Jetfeliks746:
> It still isn't working for me. I tried both uppercase and lowercase but the getIdFromUsername still return null.

Issue has been fixed!

edited line 38 in the noblox.js Node Modules to this:
return data.find((result) => result.requestedUsername === username.toLowerCase())

policetonyR (changes requested):
I was able to resolve this by changing line 50 instead of doing your method.

`return cache.wrap('IDFromName', usernames, function () {`

MilesSRC:
Closed PR, ROBLOX has fixed the issue.

PR #785, opened 2024-03-21
Fix addDeveloperProduct & checkDeveloperProductName. addGamepass feature

emsar69:
Fixed addDeveloperProduct & checkDeveloperProductName. Added 'addGamepass' function.

### DIFF
--- a/lib/games/addDeveloperProduct.js
+++ b/lib/games/addDeveloperProduct.js
@@ -24,13 +24,13 @@ const nextFunction = (jar, token, universeId, name, priceInRobux, description) =
   return checkProductName({
     universeId,
     productName: name
-  }).then((res) => {
-    if (res.Success && res.Message === 'Name available') {
+  }).then((exists) => {
+    if (!exists) {
       return http({
-        url: '//apis.roblox.com/developer-products/v1/universes/' + universeId + '/developerproducts?name=' + name + '&description=' + description + '&priceInRobux=' + priceInRobux,
+        url: `//apis.roblox.com/developer-products/v1/universes/${universeId}/developerproducts?name=${name}&description=${description}&priceInRobux=${priceInRobux}`,
         options: {
           method: 'POST',
-          jar,
+          jar: jar,
           headers: {
             'X-CSRF-TOKEN': token
           },
@@ -47,7 +47,7 @@ const nextFunction = (jar, token, universeId, name, priceInRobux, description) =
             productId: typeof res.body === 'object' ? res.body.id : JSON.parse(res.body).id
           }
         } else {
-          throw new Error('Create product failed, ' + res.statusCode + ' ' + res.statusMessage + '')
+          throw new Error(`Create product failed, ${res.statusCode} ${res.statusMessage}`)
         }
       })
     } else {
@@ -59,7 +59,7 @@ const nextFunction = (jar, token, universeId, name, priceInRobux, description) =
 exports.func = (args) => {
   const jar = args.jar
 
-  return getGeneralToken({ jar }).then((xcsrf) => {
+  return getGeneralToken({ jar: jar }).then((xcsrf) => {
     return nextFunction(jar, xcsrf, args.universeId, args.name, args.priceInRobux, args.description)
   })
 }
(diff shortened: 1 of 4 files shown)


lib/games/checkDeveloperProductName.js
```diff
@@ -23,17 +23,25 @@ exports.func = (args) => {
...
     if (res.statusCode !== 200) {
       throw new Error('You are not logged in')
     } else {
-      return JSON.parse(res.body)
+      const data = JSON.parse(res.body)
+      let exists = false
+      data.forEach(product => {
+        if(product.name == args.productName || product.id == productId){
+          exists = true
+          return
```
Neztore, line 41: This returns undefined and the typing says it returns a boolean.
Would a better approach be:
```js
for (const product of data) {
   if (...) {
       return true

}
}
return false
```?

File: lib/games/checkDeveloperProductName.js [hunk 2]
```diff
@@ -23,17 +23,25 @@ exports.func = (args) => {
   const productId = parseInt(args.productId) ? parseInt(args.productId) : 0
 
   return http({
-    url: '//www.roblox.com/places/check-developerproduct-name?universeId=' + universeId + '&developerProductId=' + productId + '&developerProductName=' + args.productName + '&_=1515792139751',
+    url: `//apis.roblox.com/developer-products/v1/universes/${universeId}/developerproducts?pageNumber=1&pageSize=2147483647`,
```
Neztore, line 26: What is the meaning of the page size here? Why is it a large arbitrary number? how does this work when the number of pages is greater than 1? Could this use a pagination wrapper

lib/games/addGamepass.js
```diff
@@ -0,0 +1,64 @@
...
+ * @param {number} priceInRobux - The price of the product.
+ * @param {string=} description - The description of the gamepass.
+ * @returns {Promise<GamepassAddResult>}
+ * @example const noblox = require("noblox.js")
+ * // Login using your cookie
+ * noblox.addGamepass(1, "A Gamepass", 100, "A cool item.")
+**/
+
+const nextFunction = (jar, token, universeId, name, priceInRobux, description) => {
+  const FormData = require('form-data');
```
Neztore, line 23: Don't require modules in functions like this.
This does not need a new dependency - I believe the new fetch API has a built in way of building these, and they are built manually elsewhere in the library.

lib/games/addGamepass.js
```diff
@@ -0,0 +1,64 @@
...
+      jar: jar,
+      body: goofy,
+      headers: {
+        ...goofy.getHeaders(),
+        'X-CSRF-TOKEN': token
+      },
+      resolveWithFullResponse: true
+    }
+  }).then((res) => {
+    console.log(res)
```
Neztore, line 43: console.log shouldn't be there

lib/games/addDeveloperProduct.js
```diff
@@ -59,7 +59,7 @@ const nextFunction = (jar, token, universeId, name, priceInRobux, description) =
 exports.func = (args) => {
   const jar = args.jar
 
-  return getGeneralToken({ jar }).then((xcsrf) => {
+  return getGeneralToken({ jar: jar }).then((xcsrf) => {
```
Neztore, line 62: Not sure what the point of this change is

Neztore (changes requested):
Thank you for your PR, I've highlighted a few issues.
Your changes to checkDeveloperProductName would be breaking changes but we intend to move to a new major version - so that isn't an issue. I agree with your API changes there; returning a boolean is the best API.

I am not comfortable adding a dependency for form data as I believe it is done manually elsewhere and there are built in APIs for doing so (it is url encoded I believe so URLSearchParams could work)

Neztore:
Adding help wanted if anyone wants to help out and resolve the above

lib/games/checkDeveloperProductName.js
Regalijan: Per my testing, Roblox does not use cursors for this endpoint. The number specified for the page size is (2^31)-1 (the largest possible signed 32bit int), which happens to be the largest page size Roblox supports. Now, whether the api will actually return that many results, I don't know.

I agree that a pagination wrapper would be preferable to this though, given this is already undocumented to start with and I don't really expect this to work in practice (but maybe somehow it does).

Noxirity:
Just seen this pr after I posted a fix for the same function https://github.com/noblox/noblox.js/pull/806

Neztore:
Closing this as it is stale. Please open a new PR if above are fixed

PR #792, opened 2024-04-15
Removed token parameter which prevented the function from respecting custom parameter values

AEW745:
Before when you would retrieve messages, it would use the default parameter values no matter what you put when you use the function. Ex. if you change the messageTab, pageSize, or pageNumber it would only use the default values. Therefore, I removed the undefined parameter to fix it. I also added the jar: jar back because it seems to work fine with it.

I also Fixed the onMessage event where it wouldn't get messages due to change in notification name and message type. I had to add the pageSize parameter and set it to 1 so it will get the latest message because before it was sending 25 messages at a time when the event is called which is also why I had to remove the token parameter in the getMessages file because it wouldn't let me set the pageSize.

I Fixed the onNotification issue that was using a deprecated signalr package and has switched over to @microsoft/signalr.
I updated the package.json file to include the change in the dependency.

### DIFF
--- a/lib/client/onNotification.js
+++ b/lib/client/onNotification.js
@@ -1,5 +1,5 @@
 // Dependencies
-const SignalR = require('signalr-client').client
+const signalR = require('@microsoft/signalr')
 const events = require('events')
 
 // Includes
@@ -36,34 +36,50 @@ exports.func = function (args) {
       args.jar = { session: args.jar }
     }
     const session = getSession({ jar: args.jar })
-    const client = new SignalR('wss://realtime.roblox.com/notifications', ['usernotificationhub'], 3, true) // wss for https
-    client.headers.Cookie = '.ROBLOSECURITY=' + session + ';'
-    client.on('usernotificationhub', 'notification', function (name, message) {
-      notifications.emit('data', name, JSON.parse(message))
-    })
-    notifications.on('close', client.end)
-    client.serviceHandlers.connectFailed = function (err) {
-      notifications.emit('error', new Error('Connection failed: ' + err.message))
-      if (retries !== -1) {
-        if (retries > max) {
-          notifications.emit('close', new Error('Max retries reached'))
-        } else {
-          setTimeout(connect, 5000, retries + 1)
+    let userNotificationConnection = null;
+
+    userNotificationConnection = new signalR.HubConnectionBuilder()
+      .withUrl('https://realtime-signalr.roblox.com/userhub', {
+        transport: signalR.HttpTransportType.WebSockets,
+        skipNegotiation: true,
+        headers: {
+          Cookie: '.ROBLOSECURITY=' + session + ';'
+        }
+      })
+      .build();
+
+      userNotificationConnection.on('notification', function(name, message) {
+        notifications.emit('data', name, JSON.parse(message))
+      })
+
+      notifications.on('close', userNotificationConnection.stop)
+
+      userNotificationConnection.disconnected = function (err) {
+        notifications.emit('error', new Error('Connection failed: ' + err.message))
+        if (retries !== -1) {
+          if (retries > max) {
+            notifications.emit('close', new Error('Max retries reached'))
+          } else {
+            setTimeout(connect, 5000, retries + 1)
+          }
         }
       }
-    }
-    client.serviceHandlers.onerror = function (err) {
-      notifications.emit('error', err)
-    }
-    client.serviceHandlers.connected = function (connection) {
-      notifications.emit('connect', connection)
-    }
-    client.serviceHandlers.reconnecting = function () {
-      setTimeout(connect, 5000, 0)
+
+      userNotificationConnection.error = function (err) {
+        notifications.emit('error', err)
+      }
+
+      userNotificationConnection.connected = function(connection) {
+        notifications.emit('connect', connection)
+      }
+
+      userNotificationConnection.reconnecting = function () {
+        setTimeout(connect, 5000, 0)
       notifications.emit('error', new Error('Lost connection, reconnecting'))
       return true // Abort reconnection
-    }
-    client.start()
+      }
+
+      userNotificationConnection.start()
   }
   connect(-1)
   return notifications
--- a/lib/privatemessages/onMessage.js
+++ b/lib/privatemessages/onMessage.js
@@ -55,7 +55,6 @@ exports.func = function (args) {
                   const message = messages[i]
                   const id = message.id
                   if (id > latest) {
-                    console.log(id)
                     latest = id
                     onMessage.emit('data', message)
                   }
--- a/lib/privatemessages/onMessage.js
+++ b/lib/privatemessages/onMessage.js
@@ -34,19 +34,20 @@ exports.func = function (args) {
   const onMessage = new events.EventEmitter()
   let waitingForRequest = false
   let latest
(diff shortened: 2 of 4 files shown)


nottisa (changes requested):
You left a comment on line # 58.

nottisa (changes requested):
You didn't include the new dependency of `@microsoft/signalr` in your package.json. Commit [aa7ba76310f3c34b3bf2d29d13ff33a00b781e17](https://github.com/noblox/noblox.js/pull/792/commits/aa7ba76310f3c34b3bf2d29d13ff33a00b781e17)

nottisa:
Additionally, those two commits don't seem to be mentioned in your PR notes.

lib/privatemessages/onMessage.js
```diff
@@ -34,26 +34,28 @@ exports.func = function (args) {
...
+              pageNumber: 0,
+              pageSize: 1
             })
               .then(function (inbox) {
                 const messages = inbox.collection
                 for (let i = messages.length - 1; i >= 0; i--) {
                   const message = messages[i]
                   const id = message.id
                   if (id > latest) {
+                    console.log(id)
```
nottisa, line 58: This should probably be removed

lib/client/onNotification.js
```diff
@@ -1,5 +1,5 @@
 // Dependencies
-const SignalR = require('signalr-client').client
+const signalR = require('@microsoft/signalr')
```
nottisa, line 2: The dependency `@microsoft/signalr` is not included in an updated package.json.

nottisa (changes requested):
Here are some possible fixes. Sorry for the previous comments. These commits I'm recommending fixes for were not mentioned in your PR comments though.

lib/privatemessages/onMessage.js
AEW745: My bad I didn’t see that when I pushed it. I will go ahead and remove it.

AEW745:
> Here are some possible fixes. Sorry for the previous comments. These commits I'm recommending fixes for were not mentioned in your PR comments though. I also noticed this commit may be a breaking change for those who have filled in a value for the token parameter.

What is the use case for the token parameter because I don’t see a need for it?

nottisa:
There's not a use case, but it's still possible that if someone was using this function and put in a throwaway value for the token parameter their code may break when they update the library. Your changes look good, I just need to validate they work which might take a couple days unfortunately.

lib/client/onNotification.js
```diff
@@ -36,34 +36,50 @@ exports.func = function (args) {
...
-    })
-    notifications.on('close', client.end)
-    client.serviceHandlers.connectFailed = function (err) {
-      notifications.emit('error', new Error('Connection failed: ' + err.message))
-      if (retries !== -1) {
-        if (retries > max) {
-          notifications.emit('close', new Error('Max retries reached'))
-        } else {
-          setTimeout(connect, 5000, retries + 1)
+    let userNotificationConnection = null;
```
Neztore, line 39: Why is this instantiated and set to null ? could it just be set directly as on line 41?

lib/privatemessages/onMessage.js
```diff
@@ -34,19 +34,20 @@ exports.func = function (args) {
   const onMessage = new events.EventEmitter()
   let waitingForRequest = false
   let latest
-  getMessages({ jar, messageTab: 'Inbox', pageNumber: 0 })
+  getMessages({ jar: jar, messageTab: 'Inbox', pageNumber: 0, pageSize: 1 })
```
Neztore, line 37: How does getMessages work if there have been multiple messages in-between calls?
I.e.
- The event polls for messages and finds none
- User receives two messages X, Y
- Which is returned? The more recent? Does this result in some messages being missed?

lib/privatemessages/onMessage.js
```diff
@@ -34,19 +34,20 @@ exports.func = function (args) {
...
     .then(function (initial) {
       latest = initial.collection[0] ? initial.collection[0].id : 0
-      const notifications = onNotification({ jar })
+      const notifications = onNotification({ jar: jar })
```
Neztore, line 40: I note in several cases you have changed the code from `{jar}`, which works, to explicitly specifying the jar value. Is there a reason for this?

Neztore (changes requested):
Just having a look and identified a few issues though I see some have been spotted already.
Thank you for the PR! I'm happy to merge once these issues have been fixed, and I have two people confirming these fixes work. The (technically) major change is not a massive concern as I intend to make the next version bump to v5.

However, as a result of this pull request some methods would be broken still (and using the old signal package). Is it possible to move those over too? If this is not something you are interested in I will open an issue - and probably not release a new version until that is done.

lib/client/onNotification.js
AEW745: It is set to null to declare it. If you test the changes I did for the realtime API you will see that it works.

lib/privatemessages/onMessage.js
AEW745: It will send the Most recent message every time a Message Notification is received.

lib/privatemessages/onMessage.js
AEW745: I just did that because it makes more sense to me when getting the jar value.

lib/privatemessages/onMessage.js
AEW745: However, I didn't change the getMessages file so if you use getMessages separately it will show the default pageSize. pageSize is only declared in the onMessage event so each time you get a new message it will send the most recent.

Neztore: I think there is potential for this implementation to miss messages and emit the same message twice.
The message notification/event will not be fired immediately (or may not be) so it is possible to get an event and then the latest message is a different one.

I.e.
1: Message/Notification received
2: Noblox starts HTTP request
3: Another message arrives
4: Roblox responds to request with the second message

Thus, a message is lost. This is likely to be an edge case and only occur in unlucky timing scenarios, but it is possible.
Is there any data supplied with the message notification that would allow the API to return the correct message for that notification? Otherwise the best approach would be to loop through and return all, storing the latest message returned (as the current API does)

AEW745: I will be doing testing over the next few days to make sure it works like it did before I made the change. When I did the change, I thought it would send every single message so I will check it out.

AEW745:
I will be adding the token parameter back because it seems like it is working with it as well. However, I won't remove the pageSize for the onMessage because removing it will cause no messages to be received as I have tested it. When multiple users send a message to the bot with the pageSize parameter set to 1 I've noticed that all messages went through, and none was missed.

AEW745:
Upon further testing adding the token parameter while not using it in the request only works with the first message and doesn't allow anymore messages to be received.

AEW745:
I will mark the comments as resolved because attempting to fix those will be a breaking change. However, leaving it how I fixed it seems to work properly.

Neztore:
> Upon further testing adding the token parameter while not using it in the request only works with the first message and doesn't allow anymore messages to be received.

Could we work out what that parameter is actually for?

Neztore (changes requested):
See above: What does token do?

Neztore (approved):

PR #793, opened 2024-04-16
Added postToTopic() for MessagingService needs

commonly-ts:
Requires user to have set their API key. The API key requires access permission `Messaging Service API` and its operation `Publish`.
image

Use case example:
```js
// JavaScript

const noblox = require("noblox.js")

await noblox.setAPIKey(process.env.API_KEY)

const universeId = 4283343407
const topicName = "ModerateUser"
const data = { targetUser: 123456789, staffMember: 1210019099, action: "Kick" }

await noblox.publishToTopic(universeId, topicName, data)
```

```lua
-- Roblox Luau

local MessagingService = game:GetService("MessagingService")
local HttpService = game:GetService("HttpService")

MessagingService:SubscribeAsync("ModerateUser", function(message)
    -- The information passed in the JavaScript is placed into `Data` and needs to be decoded
    -- Types can be used here to parse the decoded data
    local data = HttpService:JSONDecode(message.Data)
	
    print(data)
end)
```

Roblox documentation:
https://create.roblox.com/docs/reference/cloud/messaging-service
https://create.roblox.com/docs/cloud/open-cloud/usage-messaging

### DIFF
--- a/lib/games/publishToTopic.js
+++ b/lib/games/publishToTopic.js
@@ -13,8 +13,9 @@ exports.optional = ['jar']
  * @param {Object | string} data - The data to post.
  * @returns {Promise<boolean>}
  * @example const noblox = require("noblox.js")
- * const data = { targetUser: "1210019099" }
- * await noblox.publishToTopic(2152417643, "MyTopic", data)
+ * const data = { targetUser: 123456789, staffMember: 1210019099, action: "Kick" }
+ *
+ * await noblox.publishToTopic(2152417643, "ModerateUser", data)
 **/
 
 function publishToTopic (universeId, topic, data, jar) {
@@ -38,10 +39,11 @@ function publishToTopic (universeId, topic, data, jar) {
       .then(function ({ statusCode }) {
         switch (statusCode) {
           case 200: resolve(true)
-          case 400: reject(new Error(`[${statusCode}] Bad Request, check what data is being sent`))
-          case 401: reject(new Error(`[${statusCode}] Unauthorized`))
+          case 400: reject(new Error(`[${statusCode}] Bad Request, recheck the data being sent`))
+          case 401: reject(new Error(`[${statusCode}] Unauthorized | universeId: ${universeId}`))
           case 403: reject(new Error(`[${statusCode}] Forbidden, possible insufficient API key access permissions`))
           case 500: reject(new Error(`[${statusCode}] Internal server error`))
+          default: reject(new Error(`[${statusCode}] Unknown error | universeId: ${universeId} topic: ${topic}`))
         }
       }).catch(reject)
   })
--- a/lib/games/publishToTopic.js
+++ b/lib/games/publishToTopic.js
@@ -37,14 +37,17 @@ function publishToTopic (universeId, topic, data, jar) {
 
     return http(httpOpt)
       .then(function ({ statusCode }) {
-        switch (statusCode) {
-          case 200: resolve(true)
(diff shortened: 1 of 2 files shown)


lib/games/publishToTopic.js
```diff
@@ -0,0 +1,52 @@
...
+      .then(function ({ statusCode }) {
+        switch (statusCode) {
+          case 200: resolve(true)
+          case 400: reject(new Error(`[${statusCode}] Bad Request, check what data is being sent`))
+          case 401: reject(new Error(`[${statusCode}] Unauthorized`))
+          case 403: reject(new Error(`[${statusCode}] Forbidden, Possible insufficient API key access permissions`))
+          case 500: reject(new Error(`[${statusCode}] Internal server error`))
+        }
+      }).catch(reject)
+  })
```
nottisa: Error handling should probably be handled like in this file: [lib/friends/acceptFriendRequest.js](https://github.com/noblox/noblox.js/blob/master/lib/friends/acceptFriendRequest.js#L37-L49). If I'm wrong, you don't actually need the format string and can just manually enter the error codes which might be slightly more performant.

nottisa (changes requested):
Thank you for contributing! I currently don't have that much time to test. I will test sooner or later.

lib/games/publishToTopic.js
commonly-ts: I made a later commit to replace the switch case with just an if else.
I'm not sure what you mean that I don't need to format the string?

nottisa: I just mean errors should probably be reported as they come from Roblox, and not what we currently understand those errors to be. The file I linked has dynamic error handling that should pull the errors from the API instead of being hard coded in.

commonly-ts: Alright, I completely agree. I'll make a commit for the errors to be handled like they are in acceptFriendRequest.

commonly-ts: Error handling is now updated to give proper errors. The errors this API returns are very different than others so I had to spent a little more time figuring them out.

Valid API key that has any sort of invalid access permissions/scopes will return:
Error: [403] InsufficientScope Scope not authorized.

Invalid API key will return:
Error: [401] Unauthorized Invalid API Key

lib/games/publishToTopic.js
```diff
@@ -23,32 +23,32 @@ function publishToTopic (universeId, topic, data, jar) {
...
       url: `//apis.roblox.com/messaging-service/v1/universes/${universeId}/topics/${topic}`,
-      options: {
-        json: true,
-        resolveWithFullResponse: true,
-        jar,
-        method: 'POST',
-        body: { 'message': JSON.stringify(data) },
-        headers: {
-          'Content-Type': 'application/json'
+        options: {
```
commonly-ts, line 26: Had to fix all the indents to be 2 spaces because Prettier messed it all up.

nottisa:
Looks good. I'll try and test it in the next couple days.

Neztore (changes requested):
Nottisa mentioned no longer being able to contribute to the library (I assume due to other commitments), so I am reviewing.. I am happy to merge this once the typings are updated (https://github.com/noblox/noblox.js/tree/master/typings) but the way we handle open cloud may change after tomorrows meeting.

commonly-ts:
> I am happy to merge this once the typings are updated (https://github.com/noblox/noblox.js/tree/master/typings)

I have committed the change to the typings. Unsure why it also committed changes to whitespace removal 🤷

Neztore (approved):

PR #794, opened 2024-04-17
Remove threaded.js and maxThreads

### LINKED ISSUE #708: Remove threaded.js and references to maxThreads
Discovered during #703 review; [`maxThreads`](https://github.com/noblox/noblox.js/blob/5e7f572e9e25a67c83d8f130dc4ab726fc996617/settings.json#LL4C1-L5C20) is only referenced in [`threaded.js`](https://github.com/noblox/noblox.js/blob/5e7f572e9e25a67c83d8f130dc4ab726fc996617/lib/util/threaded.js) which in turn has no utilization in the library whilst being a `util` function.

`threaded.js` appears to be a helper function for fetching multiple pages at once; a decent idea but limited use case since most of the multipage endpoints we care about are cursor based and _have_ to be made sequentially

Be mindful to remove the references in `settings.json` and the associated typings/JSDocs

nottisa:
Resolves #708 in which it was brought up that:
 - [threaded.js](https://github.com/noblox/noblox.js/blob/master/lib/util/threaded.js) was not used
 - maxThreads in [settings.json](https://github.com/noblox/noblox.js/blob/master/settings.json) was only used in [threaded.js](https://github.com/noblox/noblox.js/blob/master/lib/util/threaded.js)
Refrences to [threaded.js](https://github.com/noblox/noblox.js/blob/master/lib/util/threaded.js) were removed in later commits though the file itself was never removed.


Neztore (approved):
Looks good!

PR #797, opened 2024-04-22
Return more information than "not logged in" for getConversations

### LINKED ISSUE #753: [BUG] noblox.getConversations returns not logged in error
The noblox.getConversations function call throws a not logged in error even though noblox is logged in successfully. Removing this line removes the error. 
```
console.log(await noblox.getConversations([conversationid])) // This line errors
await noblox.declineFriendRequest(data) // This line does not
```
Error:
```
Logged in as IAMAI_Bot [4589598455]
node:internal/process/promises:279
            triggerUncaughtException(err, true /* fromPromise */);
            ^

Error: You are not logged in
    at E:\\Desktop\accountgen\RoRate\node_modules\noblox.js\lib\chat\getConversations.js:31:13
    at tryCatcher (E:\\Desktop\accountgen\RoRate\node_modules\bluebird\js\release\util.js:16:23)
    at Promise._settlePromiseFromHandler (E:\\Desktop\accountgen\RoRate\node_modules\bluebird\js\release\promise.js:547:31)
    at Promise._settlePromise (E:\\Desktop\accountgen\RoRate\node_modules\bluebird\js\release\promise.js:604:18)
    at Promise._settlePromise0 (E:\\Desktop\accountgen\RoRate\node_modules\bluebird\js\release\promise.js:649:10)
    at Promise._settlePromises (E:\\Desktop\accountgen\RoRate\node_modules\bluebird\js\release\promise.js:729:18)
    at _drainQueueStep (E:\\Desktop\accountgen\RoRate\node_modules\bluebird\js\release\async.js:93:12)
    at _drainQueue (E:\\Desktop\accountgen\RoRate\node_modules\bluebird\js\release\async.js:86:9)
    at Async._drainQueues (E:\\Desktop\accountgen\RoRate\node_modules\bluebird\js\release\async.js:102:5)
    at Immediate.Async.drainQueues [as _onImmediate] (E:\\Desktop\accountgen\RoRate\node_modules\bluebird\js\release\async.js:15:14)
    at processImmediate (node:internal/timers:466:21)
PS E:\\Desktop\accountgen\RoRate>

### PR DESCRIPTION
**Pull request by Regalijan**
Currently, `getConversations()` throws the not logged in error no matter what the actual error was. Instead we should return the actual errors that Roblox returns. We should also throw in cases where one or more of the conversation IDs don't exist as to prevent any surprises.

- closes #753

At some point we should change the others as well.

### DIFF
--- a/lib/chat/getConversations.js
+++ b/lib/chat/getConversations.js
@@ -43,20 +43,6 @@ exports.func = (args) => {
         return entry
       })
 
-      const nonexistentConversations = []
-
-      for (const conversation in conversationIds) {
-        if (!response.find((res) => res.id === conversation)) {
-          nonexistentConversations.push(conversation)
-        }
-      }
-
-      if (response.length < conversationIds.length) {
-        throw new Error(
-          `The following conversations do not exist: ${nonexistentConversations.join(', ')}`
-        )
-      }
-
       return response
     }
   })
--- a/lib/chat/getConversations.js
+++ b/lib/chat/getConversations.js
@@ -43,6 +43,20 @@ exports.func = (args) => {
         return entry
       })
 
+      const nonexistentConversations = []
+
+      for (const conversation in conversationIds) {
+        if (!response.find((res) => res.id === conversation)) {
+          nonexistentConversations.push(conversation)
+        }
+      }
+
+      if (response.length < conversationIds.length) {
+        throw new Error(
+          `The following conversations do not exist: ${nonexistentConversations.join(', ')}`
+        )
+      }
+
       return response
     }
   })
--- a/lib/chat/getConversations.js
+++ b/lib/chat/getConversations.js
@@ -13,7 +13,7 @@ exports.optional = ['jar']
  * @example const noblox = require("noblox.js")
  * // Login using your cookie
  * const conversations = await noblox.getConversations([1, 2, 3])
-**/
+ **/
 
 exports.func = (args) => {
   const jar = args.jar
@@ -28,7 +28,13 @@ exports.func = (args) => {
     }
   }).then((res) => {
     if (res.statusCode !== 200) {
-      throw new Error('You are not logged in')
+      const body = JSON.parse(res.body) || {}
+      if (body.errors && body.errors.length > 0) {
+        const errors = body.errors.map((e) => {
+          return e.message
+        })
+        throw new Error(`${res.statusCode} ${errors.join(', ')}`)
+      }
     } else {
       let response = JSON.parse(res.body)
 
--- a/test/chat.test.js
+++ b/test/chat.test.js
@@ -40,10 +40,6 @@ describe('Chat Methods', () => {
     await expect(getConversations([8212952828])).resolves.not.toThrow()
   })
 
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: lib/chat/getConversations.js [hunk 1]
```diff
@@ -28,7 +28,13 @@ exports.func = (args) => {
     }
   }).then((res) => {
     if (res.statusCode !== 200) {
-      throw new Error('You are not logged in')
+      const body = JSON.parse(res.body) || {}
+      if (body.errors && body.errors.length > 0) {
+        const errors = body.errors.map((e) => {
```
Neztore, line 33: I've noticed this pattern in a few places in the library. Is a body.errors array a standard thing across the Roblox API? Could this be moved to a utility function?

lib/chat/getConversations.js
```diff
@@ -37,6 +43,20 @@ exports.func = (args) => {
...
 
+      const nonexistentConversations = []
+
+      for (const conversation in conversationIds) {
+        if (!response.find((res) => res.id === conversation)) {
+          nonexistentConversations.push(conversation)
+        }
+      }
+
+      if (response.length < conversationIds.length) {
```
Neztore, line 54: Is it better to do this error handling for the user, or to return the data with a caveat that non-existent conversations won't be in the response?
This strikes me as throwing away data unnecessarily but open to discussion

Neztore (changes requested):

lib/chat/getConversations.js
Neztore: Also I couldn't agree more with your rationale for the change: Would it be worth defining a noblox-specific error class (i.e. one that extends Error) that could contain extra info like status code, the roblox error text and an optional description for extra context that could be set by the library?
This is again a design decision so open to discussion. Either way I think this represents a semver major change so would need to form part of v5.

Regalijan: > I've noticed this pattern in a few places in the library. Is a body.errors array a standard thing across the Roblox API? Could this be moved to a utility function?

It generally is, yes. Maybe in another PR

Regalijan: > Also I couldn't agree more with your rationale for the change: Would it be worth defining a noblox-specific error class (i.e. one that extends Error) that could contain extra info like status code, the roblox error text and an optional description for extra context that could be set by the library?
> This is again a design decision so open to discussion. Either way I think this represents a semver major change so would need to form part of v5.

Yes, definitely.

lib/chat/getConversations.js
Regalijan: Honestly we could do either but it already throws if one of them exists but can't be accessed, which would mean inconsistent behavior.

Yes we are throwing away the ones that do get returned, but this behavior is undocumented and is most likely a nasty surprise to anyone that comes across it. I also struggle to think of a situation where someone would regularly hit this.

People also tend to not pay much attention to the docs so maybe we could place a warning there but I'm not sure how useful it would be.

Neztore (approved):

Neztore:
I'd like to see error handling moved to a utility and a custom error class in future; possible WG discussion.

PR #805, opened 2024-05-23
Run the linting action on pull requests

Regalijan:
Lint check currently does not run on pull requests (until merged), it would be easier than having to remember to check manually each time.

auto should not be affected as it only publishes with a release label

### DIFF
--- a/.github/workflows/npmpublish.yml
+++ b/.github/workflows/npmpublish.yml
@@ -1,6 +1,11 @@
 name: Test & Publish
 
 on:
+  pull_request:
+    types:
+      - opened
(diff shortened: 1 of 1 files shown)


Neztore:
Would it be more appropriate to create a new action which runs lint on PRs only?

Regalijan:
We could, it doesn't matter terribly since nothing would ever be published on a pr.

The other thing we could do is just exclude the publishing action from pull requests using an if conditional, thereby only having lint run (which I would prefer here as to avoid having duplicate actions)

Neztore (approved):

PR #806, opened 2024-05-27
Fix addDeveloperProducts & getDeveloperProducts

Noxirity:
Removed the initial check to see if the name is available, as `checkDeveloperProductName` is throwing a "You are not logged in" error due to the check-developerproduct-name endpoint no longer existing.

`Product with this name already exists` will still be thrown if that is the case
I also marked checkDeveloperProductName as deprecated

Updated getDeveloperProducts to use the new correct endpoint

My first PR so sorry if im doing anything wrong 😝

### DIFF
--- a/lib/games/addDeveloperProduct.js
+++ b/lib/games/addDeveloperProduct.js
@@ -1,6 +1,5 @@
 const http = require('../util/http.js').func
 const getGeneralToken = require('../util/getGeneralToken.js').func
-const checkProductName = require('./checkDeveloperProductName.js').func
 
 exports.required = ['universeId', 'name', 'priceInRobux']
 exports.optional = ['description', 'jar']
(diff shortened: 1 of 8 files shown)


Noxirity:
fixed that publishToTopic formatting because it was a bit messed up

luluwaffless:
why was this closed? its exactly the fix that i need and was about to work on LOL, you should reopen this
image

frrazer:
> why was this closed? its exactly the fix that i need and was about to work on LOL, you should reopen this image

i dont actually remember why i closed this sorry! 😩
does the code from this PR still work? i can reopen it if everything still works, i dont have time to check it myself right now

luluwaffless:
> i dont actually remember why i closed this sorry! 😩 does the code from this PR still work? i can reopen it if everything still works, i dont have time to check it myself right now

seems to work just fine, but the API im trying to use to get developer products is `//apis.roblox.com/developer-products/v2/universes/${universeId}/developerproducts` which doesn't require authentication for public games and already responds with all product information
the one the code uses is `//apis.roblox.com/developer-products/v1/universes/${universeId}/developerproducts` and `//apis.roblox.com/developer-products/v1/developer-products/${product.id}`, which DOES require authentication
some rewriting should do the deal, don't worry about it if you're not interested tho!

Issue #800, opened 2024-05-21
[FEATURE] Global error class


Regalijan:
Right now each method has its own error parser (if any). Apart from it being very repetitive, it does not work for certain error responses (i.e those not contained within an `errors` array).

I would like to move this to a class so that we're not reimplementing an error parser for every method.

Ideally, this class would:
- Handle most known error responses (i.e what the current parsers can handle, csrf errors, and certain 5xx errors that don't return a proper response body)
- Expose properties containing the status code and original response body

(If we're going to be honest this is more of an RFC)

This would probably be a v5 release target given that it could be breaking in some cases (could be pushed back).

PR #813, opened 2024-07-16
Fix configureGamePass

Regalijan:
- Use new endpoint
- Fix request dying because of non-string form values
- Don't tell Roblox it's a PNG when it isn't

Error handling could be a bit better but I'm saving that for #800

### DIFF
--- a/lib/games/configureGamePass.js
+++ b/lib/games/configureGamePass.js
@@ -35,20 +35,19 @@ function configureGamePass (gamePassId, name, description, price, icon, jar, tok
     }
 
     const httpOpt = {
-      url: '//www.roblox.com/game-pass/update',
+      url: `//apis.roblox.com/game-passes/v1/game-passes/${gamePassId}/details`,
       options: {
         method: 'POST',
         jar,
         headers: {
           'X-CSRF-TOKEN': token,
-          'Content-Type': 'multipart/form-data; boundary=----WebKitFormBoundaryKMFaNaAn4j7XeMO'
+          'Content-Type': 'multipart/form-data; boundary=-----------------------------205723423134616457963583450846'
         },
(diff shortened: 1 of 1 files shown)


lib/games/configureGamePass.js
```diff
@@ -26,29 +26,20 @@ exports.optional = ['description', 'price', 'icon', 'jar']
...
     const httpOpt = {
-      url: '//www.roblox.com/game-pass/update',
+      url: `//apis.roblox.com/game-passes/v1/game-passes/${gamePassId}/details`,
       options: {
         method: 'POST',
         jar,
         headers: {
           'X-CSRF-TOKEN': token,
-          'Content-Type': 'multipart/form-data; boundary=----WebKitFormBoundaryKMFaNaAn4j7XeMO'
+          'Content-Type': 'multipart/form-data; boundary=-----------------------------205723423134616457963583450846'
```
alanbixby, line 36: Does `boundary` need to be defined? I was under the impression this is automatically generated by `form-data`

Regalijan: Probably not, I just copied it from the example request you gave me (as the header was manually specified already)

Neztore (approved):
I'm happy but won't merge since I see there's a pending review from Alan

alanbixby (changes requested):
This needs updated typings documented; the old response is a `Promise<GamePassResponse>`, but it appears the new endpoint would be a `Promise<void>`.

This should be documented in the local JSDocs comment, `typings/jsDocs.ts`, and `typings/index.d.ts`

alanbixby (approved):
Retracting comment about types - the response is manually returning data about the endpoint, so the return type is the same. No updates to types needed.

PR #817, opened 2024-07-31
Update dependencies and minimum node version requirements

Neztore:
At present our package.json supports node v14 which is no longer supported by Node and restricts our abilities to pull in new packages. In this PR I bump our support to v18.18 as this is the minimum version for eslint, and the earliest maintained LTS (in theory we could support earlier minor versions if ignoring eslint, but given this is over a year old and that it will be a new major version, I do not see the need).

Changes made:
- Bumped minimum supported node version to v18.18
- Moved all CI/Actions to Node 20 (we've been having warnings about using an old version for a while)
- Bumped all our dependencies using yarn interactive-update
- Updated postinstall to use mjs as chalk is no longer a commonjs module in this version
- migrated to eslint v9

Future work:
- There is now a native fetch API. We should move to it.


alanbixby (approved):
Changes look visually OK.

PR #822, opened 2024-08-06
Remove old signalr and progress bar

Neztore:
Removing some unused dependencies. I had originally hoped to deprecate all use of Cheerio, but looks like it is still needed for uploadItem.

We now use standard directly for linting.

Also bumps version to v6.0.1 to account for an old publishing error, and for the breaking changes in #824

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "noblox.js",
-  "version": "5.0.1",
+  "version": "6.0.1",
   "description": "A Node.js API wrapper for Roblox.",
   "main": "lib/index.js",
   "types": "typings/index.d.ts",
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "types": "typings/index.d.ts",
   "scripts": {
     "docs": "jsdoc -c jsDocsConfig.json -r -t ./node_modules/better-docs",
-    "lint": "eslint lib/",
+    "lint": "standard lib/",
     "test": "jest",
     "postinstall": "node postinstall.mjs"
   },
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -45,17 +44,9 @@
...
-    "eslint-config-standard": "^17.1.0",
-    "eslint-plugin-import": "^2.29.1",
-    "eslint-plugin-jest": "28.6.0",
-    "eslint-plugin-n": "^17.10.1",
-    "eslint-plugin-node": "^11.0.0",
-    "eslint-plugin-promise": "^7.0.0",
     "jest": "^29.7.0",
     "jest-extended": "^4.0.2",
     "jsdoc": "4.0.3",
-    "progress": "^2.0.0",
```
alanbixby, line 58: OK - this was added in https://github.com/noblox/noblox.js/commit/ababc28d13c9c798898730d3aa29677a7220715f for
https://github.com/noblox/noblox.js/commit/95c35b03714bc0a787b22814a6219f0994710f66 and is no longer needed

package.json
```diff
@@ -25,8 +25,7 @@
     "cheerio": "^1.0.0-rc.10",
     "entities": "^5.0.0",
     "figlet": "^1.7.0",
-    "postman-request": "^2.88.1-postman.34",
-    "signalr-client": "0.0.20"
```
alanbixby, line 29: OK - was used in `onNotification` but is replaced as of https://github.com/noblox/noblox.js/commit/d626e6f49d504f3a2fbfa3cda85b999beee02cc5

package.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "noblox.js",
-  "version": "5.0.1",
+  "version": "6.0.1",
```
alanbixby, line 3: nitpick - `6.0.0` ?

package.json
```diff
@@ -1,12 +1,12 @@
...
   "description": "A Node.js API wrapper for Roblox.",
   "main": "lib/index.js",
   "types": "typings/index.d.ts",
   "scripts": {
     "docs": "jsdoc -c jsDocsConfig.json -r -t ./node_modules/better-docs",
-    "lint": "eslint lib/",
+    "lint": "standard lib/",
```
alanbixby, line 9: OK - linting rule works with eslint dev dependencies removed

package.json
alanbixby: Nevermind - 6.0.0 is a deprecated release on NPM, so `6.0.1` is needed

alanbixby (approved):
OK to merge

PR #825, opened 2024-08-17
Add getAuthenticatedUser typing

sv-du:
Seems like this was missed when adding this function


Neztore (approved):
Oops. Thank you for the PR

PR #826, opened 2024-08-22
merge

Prioq:


alanbixby:
I understand this appears to be an accidental PR; but an obligatory reminder that you shouldn't use a free proxy for authenticated API requests - it is both insecure and will cause your cookie to be invalidated.

https://devforum.roblox.com/t/psa-stop-using-roblox-proxies-roproxy-rprxyxyz-rprxy/1573256

PR #832, opened 2024-09-15
Add searchUsers method

### LINKED ISSUE #829: [REQUEST] searchUsers() function
There is no function to search for users from what I have seen. It would be nice to have and I am surprised it isn't already in noblox.

Endpoint: https://users.roblox.com/v1/users/search

Regalijan:
Closes #829


lib/users/searchUsers.js
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * ✅ Gets a list of users matching the keyword
+ * @category User
+ * @alias searchUsers
+ * @param {string} keyword - The search term to use
+ * @param {10 | 25 | 50 | 100} limit - The maximum number of matching users to return
+ * @param {string} cursor - The cursor to use when fetching the next or previous page
+ * @returns {Promise<UserSearchResult[]>}
+ * @example const noblox = require("noblox.js")
+ * // Log in using cookie
```
Neztore, line 18: pedantic, but you are not logging in using the cookie here

lib/users/searchUsers.js
```diff
@@ -0,0 +1,42 @@
...
+      keyword
+    }
+  })
+}
+
+exports.func = function (args) {
+  return searchUsers(
+    args.jar,
+    args.keyword,
+    args.limit ?? 10,
```
Neztore, line 39: Is this some kind of new-fangled syntax? Excuse my ignorance

lib/users/searchUsers.js
```diff
@@ -0,0 +1,42 @@
+// Includes
+const getPageResults = require('../util/getPageResults.js').func
+
+// Args
+exports.required = ['keyword']
+exports.optional = ['limit', 'cursor']
```
Neztore, line 6: Should jar be included as an optional arg here

Neztore (changes requested):

lib/users/searchUsers.js
Regalijan: Which part specifically?

Regalijan: Oh sorry it expanded to show much more than was intended, but yes this is new(ish).

It's node 14 and later so we should be fine here.

Neztore (approved):
thanks for your changes; ticking

Issue #833, opened 2024-09-28
Group Payout Eligibility


supercoolspy:
Add support for the users-payout-eligibility endpoint 

economy.roblox.com/v1/groups/{GroupId}/users-payout-eligibility?userIds={userid}
method: get
require: auth
res:  { [key: userId]: any }

PR #834, opened 2024-09-29
feat: Add getGroupPayoutEligibility

supercoolspy:
Solves #833
Makes it so that you can check if you can payout for a specific group


lib/groups/canPay.js
```diff
@@ -0,0 +1,50 @@
...
+ * @alias canPay
+ * @param {number} group - The id of the group.
+ * @param {number} member - The member to check payout status for.
+ * @returns {Promise<PayoutAllowedList>}
+ * @example const noblox = require("noblox.js")
+ * const groupShout = await noblox.getShout(1)
+ **/
+
+// Define
+function getShout (group, member, jar) {
```
Regalijan, line 21: Apart from the fact that this was forgotten, I'm not sure if "canPay" is a good name for this method.

Maybe `getGroupPayoutEligibility`?

Regalijan:
Also the function type is missing from the type declarations file (but please sort out the name part first).

Regalijan (changes requested):

Regalijan:
We'll need to write a test case for this too but I'll wait on that since I need to ask someone else if our test bot actually owns a group to test this with.

lib/groups/canPay.js
supercoolspy: Yeah, that makes more sense

Regalijan:
Can you sync your fork?

We made a breaking change to error handling some time back and I need to cover that before I can approve this.

PR #840, opened 2024-10-29
Split getPlayerInfo into smaller methods

### LINKED ISSUE #830: Opting out to getPlayerInfo data to reduce ratelimits
## Before posting a feature request, make sure that it hasn't been suggested before, is a feasible idea, and can benefit everyone.

**Is your feature request related to a problem? Please describe.**
While doing too many requests to getPlayerInfo, "friends.roblox.com" usually returns a 429, throwing a rejection and breaking stuff, due to the function doing [3 requests to that domain](https://github.com/noblox/noblox.js/blob/master/lib/users/getPlayerInfo.js#L24-L27) at once per getPlayerInfo call.

**Describe the solution you'd like**
I propose to do a opt-out to data I don't want to request. (A rate-limit handler would be great too but that would require much work than this I guess)
For eg., when doing tons of requests from users, I don't want to get friends or follower counts, neither username history, that would reduce processing time and avoid hitting ratelimits.

**Describe alternatives you've considered**
Another argument in getPlayerInfo where I can disable requests to "friends.roblox.com" and username history. For example:
```js
noblox.getPlayerInfo(1, {friends: false, followings: false, followers: false, usernameHistory: false})
```

**Additional context**
N/A

Regalijan:
This would resolve #830


Neztore (approved):

Neztore:
I will release a new minor version once this is merged

PR #841, opened 2024-10-29
Add RobloxAPIError class

### LINKED ISSUE #800: [FEATURE] Global error class

Regalijan:
Would resolve #800 

Draft for now because I'm not sure if I want to stop it here but I'm not sure what else to add to it (another RFC time)

### DIFF
--- a/lib/util/apiError.js
+++ b/lib/util/apiError.js
@@ -1,0 +1,16 @@
+// The data should be a response object
+
+class RobloxAPIError extends Error {
+  constructor(data) {
+    this.httpStatusCode = data.statusCode
+    this.responseBody = (function () {
+      if (typeof data.body === "string") return data.body;
+
+      try {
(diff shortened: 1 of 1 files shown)


Regalijan:
Honestly I don't really think it needs anything else feature-wise

Neztore (approved):

PR #850, opened 2024-12-13
Fix getPlayerInfo again

Regalijan:
There are three issues which this solves

- The returned response/value object was never an array to begin with and that finally stopped working at some point that I can't figure out (which is now made moot by the fact that we no longer return response objects since #840)
- We were accessing the `.body` property on the body itself since that change and I forgot to change that beforehand.
- We would throw `User does not exist` regardless of what the true error was, now the error(s) is actually logged.

EDIT: Also drops `getUsernameHistory` so people don't get ratelimited all the time.

### DIFF
--- a/lib/friends/getFollowerCount.js
+++ b/lib/friends/getFollowerCount.js
@@ -16,9 +16,9 @@ exports.required = ['userId']
 **/
 
 // Define
-exports.func = function (userId) {
+exports.func = function (args) {
   const httpOpt = {
-    url: `//friends.roblox.com/v1/users/${userId}/followers/count`,
+    url: `//friends.roblox.com/v1/users/${args.userId}/followers/count`,
     options: {
       json: true,
       method: 'GET',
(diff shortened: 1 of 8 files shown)


Neztore (changes requested):
This code isn't usable at present -we need to find the source of the 429s. If we need to drop functionality so be it

alanbixby:
Is `getUsernameHistory` the source of the 429's - I was under the impression it was the follower endpoints? 

Arguably, we may want to deprecate this function altogether, and add a `console.warn` if it 429's pointing to the new split off methods in #840 and then stage a new release.

Regalijan:
Yeah the request rate is 1/2 per minute on a clean residential ip.

This arguably should be deprecates but that wasn't brought up when this was first opened.

alanbixby (commented):
image

Receiving rate limits on follower endpoints on a residential IP within ~15 reqs/min - going to deprecate this and trigger a banner pointing to the individualized endpoints

image

alanbixby (approved):

PR #851, opened 2024-12-14
UserNotification Cloud API

commonly-ts:
Allows a notification to be sent to a user through a universe.

## Example usage
```js
// JavaScript

await noblox.setAPIKey(API_KEY!);

const universeId = 4434277335;
const targetUserId = 1210019099;
const assetId = "774d62e5-3414-b84a-89dd-77d96f1a3d33";

const parameters = {
    data: { stringValue: "Hello world!" }
};

await noblox.sendUserNotification(universeId, targetUserId, assetId, parameters);
```

Notification: 
image
image

## How it works
This pull request adds the sendUserNotification function that allows for example, a Discord bot to send a user a notification for a use case such as an application centre. A notification can be sent out to alert the user of their applications status.

Notifications are created through the Notifications tab under Engagement.
image

Custom text that is sent through the API is to be named and placed inside curly braces `{myMessage}`. To set the message, the code would look like this:
```js
const parameters = {
    myMessage: { stringValue: "This is my message." }
};
```

#### This API requires the use of an API key. The given universeId must be added to the API key and given the scope `user.user-notification:write`.
image

## Documentation
https://create.roblox.com/docs/production/promotion/experience-notifications#creating-a-notification-string
https://create.roblox.com/docs/cloud/reference/UserNotification


typings/index.d.ts
```diff
@@ -11,6 +11,7 @@ declare module "noblox.js" {
      */
     interface CookieJar {
         session?: string;
+        apiKey?: string;
```
commonly-ts, line 14: This should've been updated a while ago.

lib/games/sendUserNotification.js
```diff
@@ -0,0 +1,69 @@
...
+    return http(httpOpt)
+      .then(function (res) {
+        if (res.statusCode === 200) {
+          resolve(true)
+        } else {
+          if (typeof res.body === 'string') {
+            reject(new Error(`[${res.statusCode}] ${res.statusMessage} ${res.body}`))
+          } else {
+            const data = Object.assign(res.body)
+            reject(new Error(`[${res.statusCode}] ${data.Error} ${data.Message}`))
```
Regalijan, line 59: The error handling doesn't really work here.

V2 open cloud doesn't use the error property as far as I know (and Roblox generally follows camel case for this)

Regalijan (changes requested):
Also there's no typing added to the jsdocs

lib/games/sendUserNotification.js
Regalijan: It does catch, so it's not completely broken.

But Roblox returns a v2 error model because it's a v2 endpoint, so it comes out as `[code] undefined undefined`

The properties would be `.code` and `.message`

Regalijan:
It needs a test case too but I'll probably do that myself

lib/games/sendUserNotification.js
commonly-ts: I never got that issue with it. Have you tested it?

Neztore:
What state is this PR in? @Regalijan

Regalijan:
When I last checked it works, the problem is that I can't make a test for it until someone gets the account ready.

The api key also needs the user notification scope for that matter.

Regalijan (approved):
Looks fine to me now

PR #854, opened 2024-12-28
Fix getUserInfo

TheCakeChicken:
Updated the getUserInfo function to handle arguments correctly.

Very simple fix and a small change, however I did spend longer than I'd like to admit debugging it 😅


Regalijan:
Yeah the arg wrapping is rather unintuitive, and probably won't survive to the next major release (or potentially v8) unless someone really wants to keep it for whatever reason.

Thanks for the fix

Regalijan (approved):

PR #855, opened 2025-01-02
Migrate methods to RobloxAPIError class

Regalijan:
Currently draft because I may not be totally done but am open to help if someone is willing.

Does not include:
- Anything involving SignalR (will possibly make something else for this)
- `upload[thing]` methods since those need new endpoints
- `buy()` since it doesn't work anymore, requires multiple endpoints + 2fa, and is generally abused [eligible for removal]
- `configureItem` and `canManage` since those also need new endpoints

### DIFF
--- a/lib/games/addDeveloperProduct.js
+++ b/lib/games/addDeveloperProduct.js
@@ -1,5 +1,6 @@
 const http = require('../util/http.js').func
 const getGeneralToken = require('../util/getGeneralToken.js').func
+const RobloxAPIError = require('../util/apiError.js')
 
 exports.required = ['universeId', 'name', 'priceInRobux']
 exports.optional = ['description', 'jar']
@@ -38,8 +39,8 @@ const nextFunction = (jar, token, universeId, name, priceInRobux, description) =
         return json
       }
       throw new Error(json)
-    } catch (err) {
-      throw new Error(res.body)
+    } catch (_) {
+      throw new RobloxAPIError(res)
     }
   })
 }
--- a/lib/accountinformation/getUserSocialLinks.js
+++ b/lib/accountinformation/getUserSocialLinks.js
@@ -1,5 +1,6 @@
 // Includes
 const http = require('../util/http.js').func
+const RobloxAPIError = require('../util/apiError.js')
 
 // Args
 exports.required = ['userId']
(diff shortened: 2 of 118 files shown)


Regalijan:
Yeah at this point I'm not really sure what else to do with this.

Review time it is.

Neztore:
Wow, big PR. Thanks - will review this weekend. From a quick read over, looking at ApiError (I know it hasn't changed in this PR but you authored it originally:

```js
function getResponseBody (data) {
  if (typeof data.body === 'string') return data.body

  try {
    return JSON.stringify(data.body)
  } catch {
    throw Error('The passed response body is not a valid object')
  }
}
```



What is the behaviour here if the value passed is not a valid object? Is the actual API error returned from Roblox swallowed and the 'The passed response body is not a valid object' returned instead? Perhaps we could have the robloxApiError class as we currently do and another error class (like UnknownRobloxApiError) which exposes the raw text.

Regalijan:
> Wow, big PR. Thanks - will review this weekend. From a quick read over, looking at ApiError (I know it hasn't changed in this PR but you authored it originally:
> 
> ```js
> function getResponseBody (data) {
>   if (typeof data.body === 'string') return data.body
> 
>   try {
>     return JSON.stringify(data.body)
>   } catch {
>     throw Error('The passed response body is not a valid object')
>   }
> }
> ```
> 
> What is the behaviour here if the value passed is not a valid object? Is the actual API error returned from Roblox swallowed and the 'The passed response body is not a valid object' returned instead? Perhaps we could have the robloxApiError class as we currently do and another error class (like UnknownRobloxApiError) which exposes the raw text.

It will get swallowed. However, anything returned by Roblox should not reach this point (as all requests made through `http` come back as full responses [which have string/non-parsed bodies]). It is a failsafe against people misusing it, e.g. importing the class and passing a date object as the body.

Regalijan:
Yeah that's fine

I'll try to deal with linting tonight

lib/games/addDeveloperProduct.js
```diff
@@ -38,8 +39,8 @@ const nextFunction = (jar, token, universeId, name, priceInRobux, description) =
         return json
       }
       throw new Error(json)
```
Neztore, line 41: Sorry just been reading through and spotted this
- Would rather avoid throwing and immediately catching errors 
- This will throw a JSON parsing error as an API error, which it isn't necessarily
Could you please refactor to separate the JSON parsing error (& associated try-catch) and the API error handling?

Neztore (changes requested):

PR #856, opened 2025-02-12
Fix CSRF token retry

Regalijan:
Roblox changed the wording of the token validation error, removing the message check entirely is the better choice since there is no guarantee that they won't do it again (and was a bad idea to begin with).

### DIFF
--- a/lib/util/http.js
+++ b/lib/util/http.js
@@ -83,32 +83,22 @@ exports.func = function (args) {
   }
   return http(args.url, opt).then(function (res) {
     if (opt && opt.headers && opt.headers['X-CSRF-TOKEN']) {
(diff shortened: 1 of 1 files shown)


chaysonthelen:
Thanks for this. I am running with this PR.

alanbixby:
Won't this change cause HTTP 403 insufficient permission errors to throw an error for failure to fetch CSRF - even though the CSRF is valid?

image


Also I applied a manual edit to prevent `depth` from being double incremented - the committed code would've only had 2 attempts before erroring.

Regalijan:
Permission errors don't have a csrf token header in the response, so no.

alanbixby (approved):

PR #863, opened 2025-05-05
Request forwarding for CI

Regalijan:
Right now we cannot actually test anything requiring a cookie because of region locking (which is most of this library). My solution is to use a cloud run service with direct vpc egress (which allows us to utilize a single static ip address for all requests). This would utilize workload identity federation which avoids the need for any additional secrets (by using the oidc token provided by actions).

Some actions updates were also included (and I verified they continued to work)

### DIFF
--- a/lib/util/http.js
+++ b/lib/util/http.js
@@ -67,8 +67,8 @@ function http (url, opt) {
     const { hostname } = urlObj
 
     opt.headers["Destination-Host"] = hostname
-    opt.headers["x-serverless-authorization"] = process.env.ID_TOKEN
-    urlObj.hostname = process.env.FORWARDER_URL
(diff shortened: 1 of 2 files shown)


lib/util/http.js
```diff
@@ -62,6 +62,15 @@ function http (url, opt) {
   if (url.indexOf('http') !== 0) {
     url = 'https:' + url
   }
+  if (process?.env.CI && process.env.FORWARDER_HOSTNAME && opt.headers) {
```
Neztore, line 65: Could we add some comments explaining what these lines do?
I realise it's pretty self explanatory, but to an automated tool or uneducated person reasoning the code it may be mis-identified as a backdoor re-directing requests (and authentication tokens)

Neztore (changes requested):

Neztore (approved):

PR #865, opened 2025-06-10
Support caching the value `0` (Fixes #864)

### LINKED ISSUE #864: [BUG] Cache module does not support the value `0`
**Describe the bug**
The built-in cache module does not differentiate between `0` and `undefined`/`null` values. As a result, if the value `0` is cached, the cached value is not used.

**To Reproduce**
1. Set `"expire": true` for `"Rank"` in `settings.json`
2. Add a log statement above [this line](https://github.com/noblox/noblox.js/blob/c0a9b7a448bbb2d6fde2820fc0db6653d26d4be9/lib/cache/wrap.js#L14) in `wrap.js` to indicate when a value is added to the cache
3. Use the following test:
```js
for (let i = 0; i < 4; i++)
    await noblox.getRankInGroup(1, 1); // rank will be 0
```
4. Observe that the item `0` is fetched and added to the cache for every loop iteration

**Expected behavior**
The value `0` should be added to the cache one time, and then the cached value should be used for future calls.

**Screenshots**
Image

**Additional context**
This will affect any `getRankInGroup` call where the user is not in the group (rank 0).

mutex-lock:
Simple one-line change to fix a bug when the value `0` is cached.

### DIFF
--- a/lib/cache/addIf.js
+++ b/lib/cache/addIf.js
@@ -7,7 +7,7 @@ module.exports = function (cache, type, index, callbacks) {
   const got = get(cache, type, index)
   const item = got[0]
   const refresh = got[1]
-  if (item) {
+  if (item != null) {
     callbacks.done(item)
     if (refresh) {
       const group = cache[type]


Regalijan (approved):
looks fine to me

PR #870, opened 2025-08-07
fix: Update user agent per Roblox's request

Regalijan:
Not really a fix but not exactly a feature either


Neztore (changes requested):
See above

Neztore (approved):
Approved. Happy for you to merge once you have reviewed my changes.

PR #877, opened 2025-12-04
feat: I added some changes such as Group Bans, Group Unbans, and Join and Leave to audit logs!

AEW745:
I added Join and Leave logs to the existing Audit logs because you have to sort by Join and Leave separately as Roblox doesn't include them in the 'all' sort. I also added 2 new files such as ban and unban for the new Group bans feature.

### DIFF
--- a/lib/privatemessages/onMessage.js
+++ b/lib/privatemessages/onMessage.js
@@ -34,17 +34,17 @@ exports.func = function (args) {
   const onMessage = new events.EventEmitter()
   let waitingForRequest = false
   let latest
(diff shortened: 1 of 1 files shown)


lib/privatemessages/onMessage.js
commonly-ts: `jar: jar` is unnecessary, just do `jar`, if the parameter and variable have the same name, you can just enter `jar`

AEW745: Ok

Regalijan:
Your commit history is extremely polluted, and overwrites later modifications to the library.
On your next PR, be *very* sure you are tracking `next` instead of `master`

We also already have group ban methods, so no need for those.

